PR #5, opened 2024-05-29
[DRAFT] XP & Bling

dhvanipa:
- XP
- Bling
- Owned chests


dhvanipa:
not baking XP into core world

PR #29, opened 2025-02-11
WIP EntityId

### LINKED ISSUE #28: User defined type for `EntityId`
By using a custom type for entity ids, we get better type safety and allows us to define global helper functions for it. For example we could have things like `entityId.baseEntityId()`.

vdrg:
What do you guys think of this approach? Still missing a few files where I need to replace it.

Fixes #28


packages/world/src/EntityId.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { BaseEntity } from "./codegen/tables/BaseEntity.sol";
+
+type EntityId is bytes32;
+
+function baseEntityId(EntityId self) view returns (EntityId) {
+  EntityId base = BaseEntity._get(self);
```
dhvanipa, line 9: This would fail if it's used in the `experience` namespace right? Since its no longer root?

vdrg: ah you are right, let's figure out first if we want to keep root systems

dhvanipa (approved):
lgtm!

PR #30, opened 2025-02-11
Random Spawn

vdrg:
Tests missing


dhvanipa:
on random spawn should also check if its in an active force field. If so, dont allow it

PR #33, opened 2025-02-12
feat: add terrain system

alvrs:


packages/world/mud.config.ts
```diff
@@ -355,10 +364,10 @@ export default defineWorld({
...
-    {
-      artifactPath: "@latticexyz/world-modules/out/PuppetModule.sol/PuppetModule.json",
-      root: false,
-      args: [],
-    },
+    // {
+    //   artifactPath: "@latticexyz/world-modules/out/PuppetModule.sol/PuppetModule.json",
+    //   root: false,
+    //   args: [],
+    // },
```
alvrs, line 371: this broke the dev process trying to update the world, do we currently need it?

packages/world/src/Constants.sol
```diff
@@ -22,3 +22,6 @@ int32 constant MAX_PLAYER_INFLUENCE_HALF_WIDTH = 10;
 int32 constant MAX_PLAYER_RESPAWN_HALF_WIDTH = 10;
 
 address constant ORACLE_ADDRESS = 0xe67AcfFCe0B542540F1520a9eaD7Aa86ff31196E;
+
+int32 constant CHUNK_SIZE = 16;
+int32 constant AREA_SIZE = 512;
```
alvrs, line 27: Not used yet but the idea is an area is 512x512 (max height) and consists of many chunks. We commit to an area via a merkle root, and also already set the initial energy. Then as chunks are explored, we verify them via merkle proof to the area's merkle root.

packages/world/src/Types.sol
```diff
@@ -11,6 +11,12 @@ struct VoxelCoord {
   int32 z;
 }
 
+struct ChunkCoord {
+  int32 x;
+  int32 y;
+  int32 z;
+}
```
alvrs, line 18: intentionally creating a different type here to not get confused with the different coordinate systems (blocks vs chunks)

packages/world/src/systems/TerrainSystem.sol
```diff
@@ -0,0 +1,72 @@
...
+  function _getRelativeCoord(VoxelCoord memory coord) internal pure returns (VoxelCoord memory) {
+    return
+      VoxelCoord({
+        x: int32(uint32(mod(coord.x, CHUNK_SIZE))),
+        y: int32(uint32(mod(coord.y, CHUNK_SIZE))),
+        z: int32(uint32(mod(coord.z, CHUNK_SIZE)))
+      });
+  }
+
+  // Get the index within the chunk of a voxel coordinate
```
alvrs, line 63:
```suggestion
  // Get of a voxel coordinate within an encoded chunk
```

packages/world/mud.config.ts
dhvanipa: no can remove it, if we need it can be added back later via a script vs here

packages/world/mud.config.ts
```diff
@@ -129,6 +129,15 @@ export default defineWorld({
...
       key: ["x", "y", "z"],
     },
+    ExploredChunk: {
+      schema: {
+        x: "int32",
+        y: "int32",
+        z: "int32",
+        pointer: "address",
+      },
+      key: ["x", "y", "z"],
```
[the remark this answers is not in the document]
alvrs: Updated to use deterministic chunk deployments so we can skip the table lookup for `getBlockType`. Still want to set some table value when a chunk got explored so we can use it to update the client, but now that we don't need to track the chunk pointer here I'm tracking the address of the player who explored the chunk (might be an incentive to explore the map)

packages/world/src/systems/TerrainSystem.sol
```diff
@@ -0,0 +1,102 @@
...
+    return
+      VERSION_PADDING +
+      uint256(
+        int256(relativeCoord.x) * CHUNK_SIZE ** 2 + int256(relativeCoord.y) * CHUNK_SIZE + int256(relativeCoord.z)
+      );
+  }
+
+  /// @dev Get the salt for a chunk coordinate
+  function _getChunkSalt(ChunkCoord memory coord) internal pure returns (bytes32) {
+    return bytes32(uint256((int256(coord.x) << 64) | (int256(coord.y) << 32) | int256(coord.z)));
```
dhvanipa, line 87: Got a bit confused what it was doing here. I assume the goal here is to encode the coord as a bytes32? Is this done in this bit shifting way for gas reasons vs:
```
keccak256(abi.encodePacked(coord.x, coord.y, coord.z))
```

alvrs: Yeah correct, we just need a unique bytes32 salt per chunk coordinate, so we don't need to hash and can save some gas. Also using bitshift instead of abi.encodePacked here to save some gas. There is a test case that shows the equivalence of this operation with a different approach using abi.encodePacked

PR #34, opened 2025-02-13
Move VoxelCoord functions to a library

vdrg:
wdyt of encapsulating different types like this?


dhvanipa (approved):
very clean, I like it

PR #35, opened 2025-02-13
Energy mass initial formulas constants

dhvanipa:


packages/world/src/systems/libraries/TerrainLib.sol
```diff
@@ -15,7 +15,7 @@ library TerrainLib {
 
   /// @notice Get the terrain block type of a voxel coordinate.
   /// @dev Assumes to be called from a root system.
-  function _getBlockType(VoxelCoord memory coord) internal view returns (uint8) {
+  function _getBlockType(VoxelCoord memory coord) public view returns (uint8) {
```
vdrg, line 18: are you hitting contract size limits?

packages/world/src/systems/libraries/TransferLib.sol
```diff
@@ -53,13 +52,38 @@ library TransferLib {
...
+    VoxelCoord memory shardCoord = chestCoord.toLocalEnergyPoolShardCoord();
+    uint128 energyCost = PLAYER_TRANSFER_ENERGY_COST;
+
     EntityId forceFieldEntityId = getForceField(chestCoord);
     uint256 machineEnergyLevel = 0;
     if (forceFieldEntityId.exists()) {
+      energyCost += SMART_CHEST_ENERGY_COST;
       EnergyData memory machineData = updateMachineEnergyLevel(forceFieldEntityId);
       machineEnergyLevel = machineData.energy;
+      require(machineData.energy >= SMART_CHEST_ENERGY_COST, "Machine does not have enough energy");
```
vdrg, line 64: feels like some of this stuff (checking enough energy and decreasing from an entity) can be extracted to some util

packages/world/src/systems/libraries/TransferLib.sol
```diff
@@ -53,13 +52,38 @@ library TransferLib {
...
+      Energy._set(
+        forceFieldEntityId,
+        EnergyData({ energy: machineData.energy - SMART_CHEST_ENERGY_COST, lastUpdatedTime: uint128(block.timestamp) })
+      );
+    }
+
+    {
+      uint128 playerEnergy = Energy._getEnergy(playerEntityId);
```
vdrg, line 72: same with this, could become `playerEntityId.decreaseEnergy(amount);`

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -76,28 +76,30 @@ function removeFromInventoryCount(EntityId ownerEntityId, ObjectTypeId objectTyp
...
-    uint128 durabilityLeft = Mass._getMass(inventoryEntityId);
-    // Allow mining even if durability is exactly or less than required, then break the tool
-    require(durabilityLeft > 0, "Tool is already broken");
+    inventoryObjectTypeId = ObjectType._get(inventoryEntityId);
+    require(inventoryObjectTypeId.isTool(), "Inventory item is not a tool");
+    uint128 massLeft = Mass._getMass(inventoryEntityId);
+    require(massLeft > 0, "Tool is already broken");
+
+    // TODO: separate mine and hit?
+    // use 10% of the mass if it's greater than 0
```
vdrg, line 88:
```suggestion
    // use 10% of the mass if it's greater than 10
```

packages/world/mud.config.ts
```diff
@@ -243,6 +243,15 @@ export default defineWorld({
       },
       key: ["x", "y", "z"],
     },
+    ForceFieldMetadata: {
```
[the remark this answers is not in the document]
dhvanipa: There might not be an entity yet though. Eg you build a wall first, then place down the force field. In that case we wouldnt have tracked the mass of each block

packages/world/src/systems/libraries/TerrainLib.sol
dhvanipa: yup! when I set it to internal both build system and mine system hit limits

packages/world/src/systems/libraries/TransferLib.sol
dhvanipa: yee I added `transferEnergyFromPlayerToPool` in `EnergyUtils.sol`, curious what else could be moved

packages/world/src/systems/libraries/TransferLib.sol
dhvanipa: oh and make `PlayerEntity` a user type?

dhvanipa: Added a separate issue for this https://github.com/tenetxyz/biomes-contracts/issues/42

packages/world/src/systems/BuildSystem.sol
```diff
@@ -72,6 +75,16 @@ contract BuildSystem is System {
         BaseEntity._set(entityId, baseEntityId);
       }
     }
+    uint32 mass = ObjectTypeMetadata._getMass(objectTypeId);
+    Mass._setMass(baseEntityId, mass);
+    VoxelCoord memory shardCoord = coord.toForceFieldShardCoord();
+    ForceFieldMetadata._setTotalMassInside(
```
vdrg, line 81: mm does force field align with some other kind of shard? We might want to call it something more generic? not sure

packages/world/src/systems/libraries/TransferLib.sol
vdrg: I think this logic is similar to this for example: https://github.com/tenetxyz/biomes-contracts/pull/35/files#diff-34518cfe1e13072a67175dd059805003eb061c5b3ad4280ce3bc74f0b5cb8e39R85-R90

dhvanipa: ahh gotcha updated, added a util to `EntityId` `decreaseEnergy`

packages/world/src/systems/libraries/TransferLib.sol
dhvanipa: nvm added a util `decreaseEnergy`

packages/world/src/systems/BuildSystem.sol
dhvanipa: The force field shard is its own dimension to prevent overlaps of force fields. Renamed the variable from `shardCoord` to `forceFieldShardCoord`

packages/world/src/VoxelCoord.sol
```diff
@@ -96,11 +97,29 @@ library VoxelCoordLib {
   }
 
   function toForceFieldShardCoord(VoxelCoord memory coord) internal pure returns (VoxelCoord memory) {
-    return toShardCoord(coord, FORCE_FIELD_SHARD_DIM, true);
+    return toShardCoord(coord, FORCE_FIELD_SHARD_DIM, false);
```
dhvanipa, line 100: Force fields are 3D so we dont want to ignore Y here, but energy pools are 2D so we want to ignore Y there

packages/world/src/systems/MoveSystem.sol
```diff
@@ -4,26 +4,144 @@ pragma solidity >=0.8.24;
...
     for (uint256 i = 0; i < directions.length; i++) {
       newCoords[i] = (i == 0 ? playerCoord : newCoords[i - 1]).transform(directions[i]);
     }
 
-    MoveLib.movePlayer(playerEntityId, playerCoord, newCoords);
+    MoveLib.movePlayer(playerEntityId, playerCoord, playerEnergyData, newCoords);
+  }
+}
+
+library MoveLib {
```
dhvanipa, line 49: I moved MoveLib inside of MoveSystem since its only used here

packages/world/src/systems/libraries/TransferLib.sol
```diff
@@ -26,40 +25,53 @@ library TransferLib {
     EntityId srcEntityId,
     EntityId dstEntityId
   ) public returns (TransferCommonContext memory) {
-    (EntityId playerEntityId, ) = requireValidPlayer(msgSender);
+    (EntityId playerEntityId, , EnergyData memory playerEnergyData) = requireValidPlayer(msgSender);
 
-    EntityId baseSrcEntityId = srcEntityId.baseEntityId();
+    EntityId chestEntityId;
```
dhvanipa, line 30: Refactored this because ran into stack issues

PR #36, opened 2025-02-13
Convert chips to systems

vdrg:
- Uses fixed gas for safe chip calls (we can probably increase, it's currently a constant)
- Removes return params from chip functions
- Removes pipe systems so we can add later once we simplify transfers
- No tests for now


packages/world/src/systems/ChipSystem.sol
```diff
@@ -15,24 +18,25 @@ import { addToInventoryCount, removeFromInventoryCount } from "../utils/Inventor
...
+  function attachChipWithExtraData(EntityId entityId, ResourceId chipSystemId, bytes memory extraData) public payable {
     (EntityId playerEntityId, VoxelCoord memory playerCoord) = requireValidPlayer(_msgSender());
     VoxelCoord memory entityCoord = requireInPlayerInfluence(playerCoord, entityId);
     EntityId baseEntityId = entityId.baseEntityId();
 
     uint16 objectTypeId = ObjectType._get(baseEntityId);
-    require(Chip._getChipAddress(baseEntityId) == address(0), "Chip already attached");
-    require(chipAddress != address(0), "Invalid chip address");
+
+    // TODO: check that not already attached?
```
dhvanipa, line 38: Does the old check of reading the `Chip` table not work?

packages/world/src/systems/ChipSystem.sol
```diff
@@ -84,24 +85,26 @@ contract ChipSystem is System {
 
     addToInventoryCount(playerEntityId, PlayerObjectID, ChipObjectID, 1);
 
-    Chip._setChipAddress(baseEntityId, address(0));
+    Chip._deleteRecord(baseEntityId);
+
+    address chipAddress = baseEntityId.getChipAddress();
```
dhvanipa, line 90: if you get chipAddress after the deleteRecord, wont it just be empty?

packages/world/src/utils/callChip.sol
```diff
@@ -0,0 +1,40 @@
...
+import { Chip } from "../codegen/tables/Chip.sol";
+import { EntityId } from "../EntityId.sol";
+import { SAFE_CHIP_GAS } from "../Constants.sol";
+
+function callChip(address chipAddress, bytes memory callData) returns (bool, bytes memory) {
+  address msgSender = WorldContextConsumerLib._msgSender();
+  uint256 msgValue = WorldContextConsumerLib._msgValue();
+
+  return
+    chipAddress.call{ value: 0 }(
```
dhvanipa, line 18: With chips being systems, I thought we'd call them through the world vs the system address directly?

vdrg: We are calling it through the world (as we only should use these calls from root systems) but appending the world context manually

packages/world/src/systems/ChipSystem.sol
vdrg: thanks, we need tests lol

dhvanipa (approved):

PR #43, opened 2025-02-18
Spawn system tests

vdrg:
basic spawn system tests + common BiomesTest contract


packages/world/src/systems/SpawnSystem.sol
```diff
@@ -93,6 +101,7 @@ contract SpawnSystem is System {
   }
 
   function _spawnPlayer(uint32 playerMass, VoxelCoord memory spawnCoord) internal returns (EntityId) {
+    // TODO: check that it is revealed
```
dhvanipa, line 104: I think this already happens since if the entity exists, its already revealed, otherwise it calls `TerrainLib._getBlockType` which reverts if not revealed

packages/world/test/Spawn.t.sol
```diff
@@ -0,0 +1,184 @@
...
+
+import { massToEnergy } from "../src/utils/EnergyUtils.sol";
+import { PlayerObjectID, AirObjectID, DirtObjectID, SpawnTileObjectID } from "../src/ObjectTypeIds.sol";
+import { VoxelCoord, ChunkCoord } from "../src/Types.sol";
+import { CHUNK_SIZE, MAX_PLAYER_ENERGY } from "../src/Constants.sol";
+
+contract SpawnTest is BiomesTest {
+  function setUp() public override {
+    super.setUp();
+    ObjectTypeMetadata.setMass(PlayerObjectID, 10);
```
dhvanipa, line 23: why is a diff value set here than the default?

packages/world/test/Spawn.t.sol
```diff
@@ -0,0 +1,184 @@
...
+    ReversePosition.set(spawnTileCoord.x, spawnTileCoord.y, spawnTileCoord.z, spawnTileEntityId);
+    ObjectType.set(spawnTileEntityId, SpawnTileObjectID);
+
+    vm.prank(alice);
+    EntityId playerEntityId = world.spawn(spawnTileEntityId, spawnCoord, "");
+    assertTrue(playerEntityId.exists());
+  }
+
+  function testSpawnFailsIfNoSpawnTile() public {
+    address alice = vm.randomAddress();
```
dhvanipa, line 100: didnt know `randomAddress` was a thing, neat!

dhvanipa (approved):

packages/world/test/Spawn.t.sol
vdrg: I believe 10 is the default set in `InitPlayersSystem`

PR #47, opened 2025-02-19
Random spawn in discovered chunks

vdrg:
- Chooses between discovered chunks for random spawn
- Makes exploring a chunk a little bit more expensive
- Also implemented checks for chips
  - only private systems that can't be upgraded


packages/world/src/systems/ChipSystem.sol
```diff
@@ -37,7 +40,10 @@ contract ChipSystem is System {
 
     ObjectTypeId objectTypeId = ObjectType._get(baseEntityId);
 
-    (address chipAddress, ) = Systems._get(chipSystemId);
+    (address chipAddress, bool publicAccess) = Systems._get(chipSystemId);
+    require(!publicAccess, "Chip system must be private");
+    // TODO: do we want this?
+    require(NamespaceOwner._get(chipSystemId.getNamespaceId()) == address(0), "Chip namespace's owner not zero");
```
dhvanipa, line 46: I think we should enforce private access but not the namespace owner check yet (could comment it out but keep the check)

We ourselves will likely keep our chips upgradable for security issues right now when things are still beta

dhvanipa (approved):

PR #48, opened 2025-02-19
Multiple output recipes

vdrg:
- Support multiple outputs
- Simplified CraftSystem


packages/world/src/utils/RecipeUtils.sol
```diff
@@ -5,33 +5,57 @@ import { Recipes, RecipesData } from "../codegen/tables/Recipes.sol";
...
+function createRecipe(
+  ObjectTypeId stationObjectTypeId,
+  ObjectTypeId[] memory inputTypes,
+  uint16[] memory inputAmounts,
+  ObjectTypeId[] memory outputTypes,
+  uint16[] memory outputAmounts
+) {
+  bytes32 recipeId = hashInputs(stationObjectTypeId, inputTypes, inputAmounts);
+
+  uint16[] memory _outputTypes;
```
dhvanipa, line 25: Could you explain this bit of solidity? Not sure I understand whats happening here. Is it cuz you dont want to loop over and unwrap?

dhvanipa (approved):

packages/world/src/utils/RecipeUtils.sol
vdrg: yeah it's just casting the array without having to copy everything

PR #49, opened 2025-02-20
Players are two voxels

dhvanipa:


packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -0,0 +1,177 @@
...
+import { ObjectTypeId, AirObjectID, PlayerObjectID } from "../../ObjectTypeIds.sol";
+import { gravityApplies, inWorldBorder } from "../../Utils.sol";
+import { PLAYER_MOVE_ENERGY_COST } from "../../Constants.sol";
+import { notify, MoveNotifData } from "../../utils/NotifUtils.sol";
+import { TerrainLib } from "./TerrainLib.sol";
+import { EntityId } from "../../EntityId.sol";
+import { transferEnergyFromPlayerToPool } from "../../utils/EnergyUtils.sol";
+import { requireValidPlayer } from "../../utils/PlayerUtils.sol";
+
+library MoveLib {
```
dhvanipa, line 25: Moved `MoveLib` back into its own file as its used in multiple systems. Consolidate gravity lib function into `MoveLib` otherwise causes circular dependency

packages/world/src/utils/NotifUtils.sol
```diff
@@ -103,119 +103,119 @@ struct RevealOreNotifData {
 
 function notify(EntityId playerEntityId, BuildNotifData memory buildNotifData) {
   PlayerActionNotif._set(
-    playerEntityId.baseEntityId(),
+    playerEntityId,
```
dhvanipa, line 106: `playerEntityId` passed around anywhere should always be the base

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -56,3 +58,49 @@ function requireInPlayerInfluence(VoxelCoord memory playerCoord, EntityId entity
   requireInPlayerInfluence(playerCoord, coord);
   return coord;
 }
+
+function createPlayer(EntityId playerEntityId, VoxelCoord memory playerCoord) {
```
dhvanipa, line 62: I tried putting these two new utils in `EntityId` but causes a circular import since `VoxelCoord` already imports `EntityId`

packages/world/src/VoxelCoord.sol
```diff
@@ -322,6 +322,15 @@ library VoxelCoordLib {
   function getPlayer(VoxelCoord memory coord) internal view returns (EntityId) {
     return ReversePlayerPosition._get(coord.x, coord.y, coord.z);
   }
+
+  function removePlayer(VoxelCoord memory coord) internal {
```
vdrg, line 326: it is a bit confusing to have both remove player and delete player imo

packages/world/test/Spawn.t.sol
```diff
@@ -69,6 +69,12 @@ contract SpawnTest is BiomesTest {
     ReversePosition.set(spawnCoord.x, spawnCoord.y, spawnCoord.z, entityId);
     ObjectType.set(entityId, AirObjectID);
 
+    {
```
vdrg, line 72: maybe for later, but we could deploy an all-air chunk for tests instead of setting each air entity manually

packages/world/src/systems/LogoffSystem.sol
```diff
@@ -15,16 +15,15 @@ import { ActionType } from "../codegen/common.sol";
...
 import { getUniqueEntity } from "../Utils.sol";
-import { requireValidPlayer } from "../utils/PlayerUtils.sol";
+import { requireValidPlayer, deletePlayer } from "../utils/PlayerUtils.sol";
 import { notify, LogoffNotifData } from "../utils/NotifUtils.sol";
 
 import { EntityId } from "../EntityId.sol";
 
 contract LogoffSystem is System {
   function logoffCommon(EntityId playerEntityId, VoxelCoord memory playerCoord) internal {
     LastKnownPosition._set(playerEntityId, playerCoord.x, playerCoord.y, playerCoord.z);
```
vdrg, line 25: should setting LastKnownPosition be inside deletePlayer?

packages/world/src/systems/LogoffSystem.sol
```diff
@@ -15,16 +15,15 @@ import { ActionType } from "../codegen/common.sol";
...
-    PlayerPosition._deleteRecord(playerEntityId);
-    ReversePlayerPosition._deleteRecord(playerCoord.x, playerCoord.y, playerCoord.z);
+    deletePlayer(playerEntityId, playerCoord);
```
vdrg, line 26: ideally deletePlayer shouldn't receive a playerCoord, as we can get everything from the entity.

Wdyt of not returning the coord from `requireValidPlayer`? It is not really used in that function

packages/world/src/systems/init/InitPlayersSystem.sol
```diff
@@ -4,15 +4,31 @@ pragma solidity >=0.8.24;
...
+      ObjectTypeMetadataData({
+        stackable: 0,
+        maxInventorySlots: 36,
+        mass: 10,
+        energy: uint32(MAX_PLAYER_ENERGY),
+        canPassThrough: false
+      })
     );
+
+    int32[] memory relativePositionsX = new int32[](1);
```
vdrg, line 25: not sure how this would look, but as schemas don't change, would it be better to have fixed schemas instead (not reading from storage)?

packages/world/src/systems/LogoffSystem.sol
dhvanipa: I'm thinking of reusing `deletePlayer` on death, and in that case don't need to set it

packages/world/src/systems/LogoffSystem.sol
dhvanipa: Hmm, but `playerCoord` is used in almost all places right after the `requireValidPlayer` call, even in this logoff call need it for `LastKnownPosition`

packages/world/src/systems/init/InitPlayersSystem.sol
dhvanipa: yea that would save gas, so we'd read this from a library?

packages/world/src/VoxelCoord.sol
dhvanipa: how about in player utils: `createPlayer` and `deletePlayer`

And for `VoxelCoordLib`: `addPlayerToGrid`, `removePlayerFromGrid`

packages/world/test/Spawn.t.sol
dhvanipa: yea this one lets do in separate PR

packages/world/src/systems/LogoffSystem.sol
vdrg: right, but it feels to me (subjective) that it has a bit of semantic overload. requireValidPlayer is already doing a bunch of stuff other than "requiring it is a valid player". And due to returning the coord, we end up with having to pass the coord to `deletePlayer`, when `deletePlayer` could just receive the entityId and get the coord internally

packages/world/src/systems/init/InitPlayersSystem.sol
vdrg: could be done inline really, or have a function that returns the schemas for each object

packages/world/src/VoxelCoord.sol
vdrg: hmm wdyt of having something like entityId.removePlayerPosition, that also takes care of schema positions and Position table?

packages/world/src/systems/LogoffSystem.sol
dhvanipa: hmm playerCoord could be moved out, but then in all the places `requireValidPlayer` is called, right after we'll need to get the coord. Even in logoff, need the player coord before calling `deletePlayer` to set `LastKnownPosition`

also I like having one spot to run the things that should happen on each tx: drain player energy, set active timestamp

Maybe we should rename `requireValidPlayer` because yea its doing more than "requiring it is a valid player". How about like `requireValidPlayer` only does the checks, and then another util `getPlayerInfo` calls `requireValidPlayer` and in addition does the other 3 things: 1) player coord, 2) update energy, 3) set `PlayerActivity`

vdrg (approved):

packages/world/src/systems/LogoffSystem.sol
vdrg: I prefer just keeping `requireValidPlayer` over that for now,  if we have something like `getPlayerInfo` I would expect it not to modify the state

vdrg: I understand the idea of grouping stuff that is usually done together, but sometimes that obscures too much what the code is doing, but we can leave for now and revisit later :)

PR #50, opened 2025-02-20
Ore commits/respawn/mining/burning

vdrg:
Current design:
- Defined constants for maximum amount of each ore, should ideally match the terrain amounts
- OreCommitments are set per chunk, they last for 256 blocks and commit to the next block
- MinedOreCount keeps track of how many of each ore has been mined
- TotalMinedOreCount keeps track of the total amount of ores that have been mined
- MinedOrePosition keeps an indexed array of the coordinates of ores have been mined
- Ore amounts for each tool are determined in the getOreAmount function
- When an ore is mined, its coordinate is added to the MinedOre tables. When respawning ores, we randomly choose from one of these coordinates.


packages/world/mud.config.ts
```diff
@@ -48,6 +48,21 @@ export default defineWorld({
       },
       key: ["objectTypeId"],
     },
+    ObjectTypeOreAmount: {
+      schema: {
+        objectTypeId: "ObjectTypeId",
+        oreTypeId: "ObjectTypeId",
+        amount: "uint16",
```
dhvanipa, line 55: Max uint16 is 65,535, maybe should be bigger?

I think there'll be more than these in the world, or can add a todo and change later once we have the exact count from the map

vdrg: this is just to track how many ores should be burned for that objectTypeId

dhvanipa: ahh right

packages/world/src/systems/OreSystem.sol
```diff
@@ -4,87 +4,84 @@ pragma solidity >=0.8.24;
...
-  function initiateOreReveal(VoxelCoord memory coord) public {
-    require(inWorldBorder(coord), "Cannot reveal ore outside world border");
-
-    (EntityId playerEntityId, VoxelCoord memory playerCoord, ) = requireValidPlayer(_msgSender());
-    requireInPlayerInfluence(playerCoord, coord);
-    if (TerrainCommitment._getBlockNumber(coord.x, coord.y, coord.z) != 0) {
-      revealOre(coord);
-      return;
-    }
+  function oreChunkCommit(ChunkCoord memory chunkCoord) public {
```
dhvanipa, line 45: why are `OreCommitments` per chunk coord vs per the ore coord?

packages/world/src/systems/MineSystem.sol
```diff
@@ -28,90 +33,160 @@ import { getObjectTypeSchema } from "../utils/ObjectTypeUtils.sol";
...
 import { EntityId } from "../EntityId.sol";
 import { PLAYER_MINE_ENERGY_COST } from "../Constants.sol";
+import { ChunkCoord } from "../Types.sol";
+import { COMMIT_EXPIRY_BLOCKS, MAX_COAL, MAX_SILVER, MAX_GOLD, MAX_DIAMOND, MAX_NEPTUNIUM } from "../Constants.sol";
+
+library MineLib {
+  function mineRandomOre(VoxelCoord memory coord) public returns (ObjectTypeId) {
+    ChunkCoord memory chunkCoord = coord.toChunkCoord();
+    uint256 blockNum = OreCommitment._get(chunkCoord.x, chunkCoord.y, chunkCoord.z);
+    require(blockNum > block.number - COMMIT_EXPIRY_BLOCKS, "Ore commitment expired");
```
dhvanipa, line 43: What if some tx fails and the user gets in this state? They cant mine any ores right

Maybe we should add a way to clear commitments?

packages/world/src/systems/OreSystem.sol
vdrg: We could do it per ore, but I think it makes sense making it per chunk while exploring, having it per ore doesn't bring any benefits imo. I just found an issue though, I'm only using the block randomness when mining, not the block coord

packages/world/src/Constants.sol
```diff
@@ -25,13 +25,16 @@ int32 constant MAX_PLAYER_INFLUENCE_HALF_WIDTH = 10;
...
+uint256 constant SPAWN_BLOCK_RANGE = 10;
 
 int32 constant CHUNK_SIZE = 16;
 int32 constant AREA_SIZE = 512;
 
 uint256 constant SAFE_CHIP_GAS = 1_000_000;
 
-address constant ORACLE_ADDRESS = 0xe67AcfFCe0B542540F1520a9eaD7Aa86ff31196E;
+uint256 constant COMMIT_EXPIRY_BLOCKS = 256;
+int32 constant COMMIT_HALF_WIDTH = 2;
```
dhvanipa, line 36: Could this constant include `CHUNK` so we know its 2 chunks not 2 voxels?

packages/world/src/systems/MineSystem.sol
```diff
@@ -28,82 +33,171 @@ import { getObjectTypeSchema } from "../utils/ObjectTypeUtils.sol";
...
-    require(entityId.getChipAddress() == address(0), "Cannot mine a chipped block");
-    EnergyData memory machineData = updateMachineEnergyLevel(entityId);
-    require(machineData.energy == 0, "Cannot mine a machine that has energy");
+  function _removeBlock(EntityId entityId, VoxelCoord memory coord) internal {
+    ObjectType._set(entityId, AirObjectID);
 
-    return (entityId, mineObjectTypeId);
+    VoxelCoord memory aboveCoord = VoxelCoord(coord.x, coord.y + 1, coord.z);
+    EntityId aboveEntityId = aboveCoord.getPlayer();
+    if (aboveEntityId.exists()) {
```
dhvanipa, line 148: why was the `&& aboveEntityId.isBaseEntity()` check removed?

I think we need that since only then would gravity apply

dhvanipa (approved):

packages/world/src/systems/MineSystem.sol
vdrg: is it possible for the above player entity not to be the base entity? That would mean the top entity (not base) is directly on top of a block, which from my understanding is not possible

dhvanipa: true I guess this is not possible cuz players are on top of each other. Was more thinking if we add another type of body like a drone that is 2 wide. But hmm yea dont need the extra read then!

vdrg: sg will add a comment for now

vdrg: actually, if we expect to have other movable objects (being players or others) shouldn't that table be called differently?

PR #53, opened 2025-02-21
Add build tests

dhvanipa:
- Refactored BuildSystem to match the style of MineSystem
- Fixed bug in move where jump builds wouldnt work
- Fixed bug in mine where it would set mass for non-base entities and allow non-pass through blocks to be built where players where
- Replaced assertTrue's with assertEq where applicable


packages/world/src/VoxelCoord.sol
```diff
@@ -315,13 +317,12 @@ library VoxelCoordLib {
       Position._set(entityId, coord.x, coord.y, coord.z);
       ReversePosition._set(coord.x, coord.y, coord.z, entityId);
       ObjectType._set(entityId, objectTypeId);
-
-      Mass._setMass(entityId, ObjectTypeMetadata._getMass(objectTypeId));
```
dhvanipa, line 319: I moved this mass check outside since this function is used for non-base entities as well. Thats why also made it return a bool indicating its this code path

packages/world/src/VoxelCoord.sol
```diff
@@ -332,6 +333,27 @@ library VoxelCoordLib {
     PlayerPosition._set(playerEntityId, coord.x, coord.y, coord.z);
     ReversePlayerPosition._set(coord.x, coord.y, coord.z, playerEntityId);
   }
+
+  /// @dev Get relative schema coords, including base coord
+  function getRelativeCoords(
```
dhvanipa, line 338: Moved here from MineLib cuz its reused in Build

packages/world/src/systems/BuildSystem.sol
```diff
@@ -30,89 +30,86 @@ import { MoveLib } from "./libraries/MoveLib.sol";
 import { EntityId } from "../EntityId.sol";
 
 contract BuildSystem is System {
-  function buildObjectAtCoord(ObjectTypeId objectTypeId, VoxelCoord memory coord) internal returns (EntityId) {
+  function _addBlock(ObjectTypeId buildObjectTypeId, VoxelCoord memory coord) internal returns (EntityId) {
```
dhvanipa, line 33: renamed to match what's in MineSystem --> `_removeBlock`

packages/world/src/systems/BuildSystem.sol
```diff
@@ -30,89 +30,86 @@ import { MoveLib } from "./libraries/MoveLib.sol";
...
-    ObjectType._set(terrainEntityId, objectTypeId);
+    ObjectType._set(terrainEntityId, buildObjectTypeId);
 
     return terrainEntityId;
   }
 
   function buildWithExtraData(
-    ObjectTypeId objectTypeId,
-    VoxelCoord memory coord,
+    ObjectTypeId buildObjectTypeId,
```
dhvanipa, line 51: I renamed these to be more explicit, was easier to read the code

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -158,7 +168,7 @@ library MoveLib {
       return false;
     }
 
-    (, ObjectTypeId belowObjectTypeId) = belowCoord.getOrCreateEntity();
+    (, ObjectTypeId belowObjectTypeId) = belowCoord.getEntity();
```
dhvanipa, line 171: in spawns, moves I realized we dont need to create the air entity, its unnecessary gas. But forgot to update gravity which is also a move, updated here

packages/world/src/systems/MineSystem.sol
```diff
@@ -175,32 +157,33 @@ contract MineSystem is System {
...
 
-    uint128 finalMass = MineLib.processMassReduction(playerEntityId, baseEntityId);
-
     // First coord will be the base coord, the rest is relative schema coords
-    VoxelCoord[] memory coords = _getRelativeCoords(mineObjectTypeId, baseCoord);
+    VoxelCoord[] memory coords = baseCoord.getRelativeCoords(mineObjectTypeId);
 
-    if (finalMass == 0) {
-      Mass._deleteRecord(baseEntityId);
+    {
```
dhvanipa, line 163: moved this into `{}` cuz of stack issues

packages/world/src/VoxelCoord.sol
vdrg: I feel we should rethink this function, it is hard to reason about what it is doing and what's the bool value for

vdrg (approved):
lgtm, will try to think how we can clean the getOrCreateEntity function a bit

packages/world/src/VoxelCoord.sol
dhvanipa: hmm the way I'm thinking about it is anytime in any system we have a coordinate and want the entity + type there, we call this function. If it doesnt exist, we call terrain system and create it.

Do you think a simple rename of the function to something better like `getOrCreateGridEntity` is enough or are you thinking of something else?

PR #54, opened 2025-02-21
Mine ore test and terrain test util

vdrg:


packages/world/src/systems/MineSystem.sol
```diff
@@ -103,7 +95,7 @@ library MineLib {
     }
 
     ObjectTypeId ore = ores[oreIndex];
-    MinedOreCount._set(ore, mined[oreIndex] + 1);
+    MinedOreCount._set(ore, max[oreIndex] - remaining[oreIndex] + 1);
```
dhvanipa, line 98: why is this not `MinedOreCount._get(ore) + 1`?

vdrg: didn't want to do a storage read

dhvanipa: ahhh right I see

dhvanipa (approved):

PR #60, opened 2025-02-25
Bed System + Orientation

vdrg:
WIP

- Keeps track of a `bedEntityId` in `PlayerStatus`
- Energy drain rates are now tracked in the Energy table
  - On sleep, the player's drain rate is decreased by the same amount the forcefield's drain rate is increased
- Energy table also tracks accumulated depletion time (maybe we should move it to a machine only table, as for players there's no use for this)
  - This is needed to be able to deplete a sleeping player's energy if the forcefield is depleted
- Remove login/logoff stuff
- Add orientation when building

TODO:

- [x] Figure out what happens when a bed is mined with a sleeping player
  - [x] Inventory is transferred between player<>bed on wakeup/sleep 
  - [x] What does it mean to kill a player?
- [x] Figure out what happens if force field runs out of energy
- [x] Figure out what happens if player runs out of energy (should wakeup revert? should we include a function to kill the player?)
- [ ] Tests


packages/world/mud.config.ts
```diff
@@ -17,8 +17,6 @@ export default defineWorld({
       "Equip",
       "Unequip",
       "Spawn",
-      "Login",
```
dhvanipa, line 20: Should we add sleep and wakeup notifs? Seems likely a quest will use them

packages/world/src/VoxelCoord.sol
```diff
@@ -317,12 +313,15 @@ library VoxelCoordLib {
       Position._set(entityId, coord.x, coord.y, coord.z);
       ReversePosition._set(coord.x, coord.y, coord.z, entityId);
       ObjectType._set(entityId, objectTypeId);
+      uint32 mass = ObjectTypeMetadata._getMass(objectTypeId);
+      if (mass > 0) {
```
dhvanipa, line 317: Won't this set the mass for relative entity ids when it shouldn't? Since this is being called in `BuildSystem`

packages/world/src/systems/MachineSystem.sol
```diff
@@ -32,9 +33,18 @@ contract MachineSystem is System {
     ObjectTypeId objectTypeId = ObjectType._get(baseEntityId);
     require(objectTypeId == ForceFieldObjectID, "Invalid object type");
     EnergyData memory machineData = updateMachineEnergyLevel(baseEntityId);
+
     uint128 newEnergyLevel = machineData.energy + (uint128(numBattery) * 10);
 
-    Energy._set(baseEntityId, EnergyData({ lastUpdatedTime: uint128(block.timestamp), energy: newEnergyLevel }));
+    Energy._set(
```
dhvanipa, line 39: could we use the `setEnergy` util here since we're not changing drain rate or accDepletedTime?

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -29,9 +33,10 @@ function requireValidPlayer(address player) returns (EntityId, VoxelCoord memory
   checkWorldStatus();
   EntityId playerEntityId = Player._get(player);
   require(playerEntityId.exists(), "Player does not exist");
-  require(!PlayerStatus._getIsLoggedOff(playerEntityId), "Player isn't logged in");
   VoxelCoord memory playerCoord = PlayerPosition._get(playerEntityId).toVoxelCoord();
   EnergyData memory playerEnergyData = updatePlayerEnergyLevel(playerEntityId);
+  require(playerEnergyData.energy != 0, "Player is dead");
```
dhvanipa, line 38: nit: should this be > 0, since it cant be negative

packages/world/src/systems/BedSystem.sol
```diff
@@ -0,0 +1,167 @@
...
+    return (machineData, playerData);
+  }
+}
+
+contract BedSystem is System {
+  using VoxelCoordLib for *;
+
+  // TODO: should this accept the coordinate of an air block to transfer inventory to?
+  function removeDeadPlayerFromBed(EntityId playerEntityId, VoxelCoord memory dropCoord) public {
+    EntityId bedEntityId = PlayerStatus._getBedEntityId(playerEntityId);
```
dhvanipa, line 63: Missing a `checkWorldStatus();` check here

packages/world/src/VoxelCoord.sol
dhvanipa: I think there's actually a test that checks that mass is only set on base entity id, so it should be failing

vdrg: hmm true, but I don't think we always get the base entity when checking the object type id, so it might be safe to do this?

vdrg: wait I don't think it does, as this is only for terrain entities no?

dhvanipa: yee you're right only terrain entities, so good for now then cuz we dont support non-voxel terrain blocks

PR #64, opened 2025-02-27
Add drained energy to local pool for players and machines

dhvanipa:
- Energy was being drained but not added to the local pool
- Removed the threshold as its complicating things
- Added tests


packages/world/src/utils/EnergyUtils.sol
```diff
@@ -20,58 +23,50 @@ function energyToMass(uint128 energy) pure returns (uint128) {
 
 function getLatestEnergyData(
   EntityId entityId,
-  uint128 drainEnergyThreshold,
   uint128 drainInterval,
   uint128 drainRate
-) view returns (EnergyData memory) {
+) view returns (EnergyData memory, uint128) {
```
dhvanipa, line 28: This is a view function so cant update the energy here

PR #71, opened 2025-02-28
Add craft tests

dhvanipa:
- Fixed bug where recipes with the same inputs but multiple outputs werent supported (was using the last set output id)
- Added a separate method for crafting recipes that don't require a station
- Added a couple of of test utils for inventory and energy checks that was being duplicated in all the tests


packages/world/src/systems/CraftSystem.sol
```diff
@@ -20,38 +20,33 @@ import { addToInventoryCount, removeFromInventoryCount, removeAnyFromInventoryCo
...
 contract CraftSystem is System {
-  function craft(EntityId stationEntityId, ObjectTypeId[] memory inputTypes, uint16[] memory inputAmounts) public {
-    require(inputTypes.length > 0, "Recipe not found");
+  function craftWithStation(bytes32 recipeId, EntityId stationEntityId) public {
+    RecipesData memory recipeData = Recipes._get(recipeId);
+    require(recipeData.inputTypes.length > 0, "Recipe not found");
 
     (EntityId playerEntityId, VoxelCoord memory playerCoord, ) = requireValidPlayer(_msgSender());
-
-    requireInPlayerInfluence(playerCoord, stationEntityId);
```
dhvanipa, line 33: requireInPlayerInfluence was incorrectly checking in cases where there is no station, so moved that logic inside

packages/world/mud.config.ts
```diff
@@ -52,6 +52,9 @@ export default defineWorld({
     Recipes: {
       schema: {
         recipeId: "bytes32",
+        stationTypeId: "ObjectTypeId",
```
dhvanipa, line 55: This is more storage than before but is a one time thing

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -87,7 +87,7 @@ function removeAnyFromInventoryCount(EntityId playerEntityId, ObjectTypeId objec
       remaining -= spend;
     }
   }
-  require(remaining == 0, "Not enough objects");
+  require(remaining == 0, "Not enough objects in the inventory");
```
dhvanipa, line 90: Made this string match the other place we do a similar check

packages/world/src/systems/CraftSystem.sol
```diff
@@ -20,38 +20,33 @@ import { addToInventoryCount, removeFromInventoryCount, removeAnyFromInventoryCo
...
-    EntityId baseStationEntityId = stationEntityId.baseEntityId();
-
-    ObjectTypeId stationObjectTypeId = ObjectType._get(baseStationEntityId);
-
-    bytes32 recipeId = hashInputs(stationObjectTypeId, inputTypes, inputAmounts);
-
-    RecipesData memory recipeData = Recipes._get(recipeId);
+    if (!recipeData.stationTypeId.isNull()) {
+      require(stationEntityId.exists(), "This recipe requires a station");
+      require(ObjectType._get(stationEntityId.baseEntityId()) == recipeData.stationTypeId, "Invalid station");
```
vdrg, line 34: is it required to use the baseEntityId? as we are setting it for all entities

packages/world/src/utils/InventoryUtils.sol
vdrg: do you think eventually we should move to custom errors? should also help a bit with contract size

vdrg: not needed for now though, can be changed later

packages/world/test/BiomesTest.sol
```diff
@@ -286,4 +292,61 @@ abstract contract BiomesTest is MudTest, GasReporter {
...
+    EntityId forceFieldEntityId = ForceField.get(
+      forceFieldShardCoord.x,
+      forceFieldShardCoord.y,
+      forceFieldShardCoord.z
+    );
+    snapshot.forceFieldEnergy = forceFieldEntityId.exists() ? Energy.getEnergy(forceFieldEntityId) : 0;
+    return snapshot;
+  }
+
+  function assertEnergyFlowedFromPlayerToLocalPool(
```
vdrg, line 343: nice! should we move these to something like "BiomesAssertions.sol", and make BiomesTest inherit from it?

vdrg: also we should figure out when we add stuff to the base test contract vs TestUtils lib

vdrg (approved):
lgtm! just a few comments

packages/world/src/utils/InventoryUtils.sol
dhvanipa: Seems like a good idea, lets create a separate issue to track it?

packages/world/src/systems/CraftSystem.sol
dhvanipa: true! removed it

packages/world/test/BiomesTest.sol
dhvanipa: moved into `BiomesAssertions`

PR #75, opened 2025-03-03
Vec3 Storage

vdrg:


packages/world/src/systems/libraries/TerrainLib.sol
```diff
@@ -2,34 +2,32 @@
 pragma solidity >=0.8.24;
 
 import { WorldContextConsumerLib } from "@latticexyz/world/src/WorldContext.sol";
-import { VoxelCoord, VoxelCoordLib } from "../../VoxelCoord.sol";
-import { ChunkCoord } from "../../Types.sol";
 import { SSTORE2 } from "../../utils/SSTORE2.sol";
 import { CHUNK_SIZE } from "../../Constants.sol";
-import { floorDiv, mod } from "../../utils/MathUtils.sol";
+import { mod } from "../../utils/MathUtils.sol";
+import { Vec3, vec3 } from "../../Vec3.sol";
```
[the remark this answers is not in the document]
vdrg: yeah exactly, we could rename it to something else though

dhvanipa (approved):

packages/world/src/Utils.sol
```diff
@@ -13,14 +13,14 @@ function checkWorldStatus() view {
...
-function inWorldBorder(VoxelCoord memory coord) pure returns (bool) {
+function inWorldBorder(Vec3 coord) pure returns (bool) {
   return
-    coord.x >= WORLD_BORDER_LOW_X &&
-    coord.x <= WORLD_BORDER_HIGH_X &&
-    coord.y >= WORLD_BORDER_LOW_Y &&
-    coord.y <= WORLD_BORDER_HIGH_Y &&
-    coord.z >= WORLD_BORDER_LOW_Z &&
-    coord.z <= WORLD_BORDER_HIGH_Z;
+    coord.x() >= WORLD_BORDER_LOW_X &&
```
dhvanipa, line 18: If we end up doing this in more places in the future could make this a util in coord, and just pass in WORLD_BORDER_LOW and WORLD_BORDER_HIGH cuz we can inline define them now I think cuz Vec3 is not a struct

packages/world/test/BiomesAssertions.sol
```diff
@@ -103,4 +93,12 @@ abstract contract BiomesAssertions is MudTest, GasReporter {
     uint128 localPoolEnergyGained = afterEnergyDataSnapshot.localPoolEnergy - beforeEnergyDataSnapshot.localPoolEnergy;
     assertEq(localPoolEnergyGained, playerEnergyLost, "Local pool energy did not gain energy");
   }
+
+  function assertEq(Vec3 a, Vec3 b, string memory err) internal pure {
```
dhvanipa, line 97: smart

packages/world/src/Vec3.sol
```diff
@@ -0,0 +1,258 @@
...
+}
+
+function sub(Vec3 a, Vec3 b) pure returns (Vec3) {
+  return vec3(a.x() - b.x(), a.y() - b.y(), a.z() - b.z());
+}
+
+function getDirectionVector(Direction direction) pure returns (Vec3) {
+  if (direction == Direction.PositiveX) return vec3(1, 0, 0);
+  if (direction == Direction.NegativeX) return vec3(-1, 0, 0);
+  if (direction == Direction.PositiveY) return vec3(0, 1, 0);
```
alvrs, line 36: not sure how often this is used but would it be worth pre-computing these instead of encoding on the fly?

packages/world/src/Vec3.sol
alvrs, line 1: :chefkiss:

packages/world/src/Utils.sol
vdrg: ah I wanted to talk about that, imo things related to other entities we probably want to keep outside vec3, as it starts becoming too coupled with everything else. What we can do for better syntax is to instead wrap the existing utils into libraries, and then use stuff like "using PlayerUtils for *" if necessary

packages/world/src/utils/Vec3Storage.sol
```diff
@@ -0,0 +1,439 @@
...
+import { EntityId } from "../EntityId.sol";
+import { Vec3 } from "../Vec3.sol";
+
+/// @dev Library to get and set Vec3s in tables. It only support schemas of <single key> -> Vec3 and Vec3 -> <single value>
+library Vec3Storage {
+  function get(ResourceId tableId, FieldLayout fieldLayout, bytes32 key) internal view returns (Vec3 output) {
+    (bytes memory _staticData, , ) = StoreSwitch.getRecord(tableId, _encodeKeyTuple(key), fieldLayout);
+
+    assembly {
+      output := mload(add(_staticData, 0xc))
```
alvrs, line 32: why are you adding 12 here? 🤔

packages/world/src/Vec3.sol
vdrg: yeah we can definitely do that, I was just lazy, but not used that much

packages/world/test/Bed.t.sol
```diff
@@ -120,9 +114,9 @@ contract BedTest is BiomesTest {
   }
 
   function testSleepFailsIfNotInPlayerInfluence() public {
-    (address alice, , VoxelCoord memory coord) = setupAirChunkWithPlayer();
+    (address alice, , Vec3 coord) = setupAirChunkWithPlayer();
 
-    VoxelCoord memory bedCoord = VoxelCoord(coord.x - 500, coord.y, coord.z);
+    Vec3 bedCoord = coord + vec3(500, 0, 0);
```
alvrs, line 119: intentional that `-500` turned into `+500`?

alvrs (approved):

packages/world/test/Bed.t.sol
vdrg: it's the same for this test, we just want to be far away, we could probably improve it by setting the min distance required for it to be out of range

packages/world/src/utils/Vec3Storage.sol
vdrg:
```
➜ bytes memory t = abi.encodePacked(uint96(123));
➜ t
Type: dynamic bytes
├ Hex (Memory):
├─ Length ([0x00:0x20]): 0x000000000000000000000000000000000000000000000000000000000000000c
├─ Contents ([0x20:..]): 0x00000000000000000000007b0000000000000000000000000000000000000000
├ Hex (Tuple Encoded):
├─ Pointer ([0x00:0x20]): 0x0000000000000000000000000000000000000000000000000000000000000020
├─ Length ([0x20:0x40]): 0x000000000000000000000000000000000000000000000000000000000000000c
└─ Contents ([0x40:..]): 0x00000000000000000000007b0000000000000000000000000000000000000000

➜ uint96 c;

➜ assembly { c := mload(t) }
➜ c
Type: uint96
├ Hex: 0x
├ Hex (full word): 0x000000000000000000000000000000000000000000000000000000000000000c
└ Decimal: 12

➜ assembly { c := mload(add(t, 0xc)) }
➜ c
Type: uint96
├ Hex: 0x
├ Hex (full word): 0x000000000000000000000000000000000000000000000000000000000000007b
└ Decimal: 123
```

vdrg: in other words it is so that the content is right aligned so we can cast it directly to uint96

alvrs:
Ahhh got it, nice, maybe let’s add a quick comment about that. I was
confused because the 32 byte slice we‘re reading would also include the
array length, but it’s cut off when it’s casted to uint96 right ?

PR #79, opened 2025-03-05
Random spawn client

dhvanipa:


packages/world/src/systems/SpawnSystem.sol
```diff
@@ -60,7 +74,7 @@ contract SpawnSystem is System {
     int32 relativeX = int32(int256(posRand % chunkSize));
     int32 relativeZ = int32(int256((posRand / chunkSize) % chunkSize));
 
-    return vec3(chunkWorldX + relativeX, y, chunkWorldZ + relativeZ);
+    return vec3(chunkWorldX + relativeX, chunkWorldY, chunkWorldZ + relativeZ);
```
dhvanipa, line 77: Changed this to return the chunk coord y so in the client we know which chunk the x and z are from which is already explored

PR #83, opened 2025-03-05
Expandable forcefields

vdrg:
TODO:
- More tests


packages/world/src/systems/MineSystem.sol
```diff
@@ -201,7 +233,12 @@ contract MineSystem is System {
       MineNotifData({ mineEntityId: baseEntityId, mineCoord: coord, mineObjectTypeId: mineObjectTypeId })
     );
 
-    ForceFieldLib.requireMinesAllowed(playerEntityId, baseEntityId, mineObjectTypeId, coords, extraData);
+    // TODO: not sure about this
+    if (mineObjectTypeId == ObjectTypes.ForceField && finalMass == 0) {
+      destroyForceField(baseEntityId);
```
dhvanipa, line 238: we still need to call the force field to check if mining is allowed otherwise anyone could mine the force field

packages/world/src/systems/admin/OracleSystem.sol
```diff
@@ -10,14 +10,14 @@ import { Vec3 } from "../../Vec3.sol";
 import { PHYSICS_ORACLE_ADDRESS } from "../../Constants.sol";
 
 contract OracleSystem is System {
-  function oracleSetInitialEnergyPool(Vec3 shardCoord, uint128 energy) public {
+  function oracleSetInitialEnergyPool(Vec3 fragmentCoord, uint128 energy) public {
```
dhvanipa, line 13: Energy pools are per shard, cuz they're 2d, fragment is 3d so this shouldnt be renamed

packages/world/src/utils/ForceFieldUtils.sol
```diff
@@ -3,21 +3,127 @@ pragma solidity >=0.8.24;
...
+  return fragmentData.entityId;
 }
 
-function destroyForceField(EntityId forceFieldEntityId, Vec3 coord) {
-  Vec3 shardCoord = coord.toForceFieldShardCoord();
-  ForceField._deleteRecord(shardCoord);
+/**
+ * @dev Destroy a forcefield
+ */
+function destroyForceField(EntityId forceFieldEntityId) {
```
dhvanipa, line 126: Can we add a comment here on how we dont clean up fragments

packages/world/src/systems/MineSystem.sol
dhvanipa: ideally we should add a test for building and mining a force field

dhvanipa (approved):

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -63,6 +63,27 @@ function requireInPlayerInfluence(Vec3 playerCoord, EntityId entityId) view retu
...
+
+  int32 range = FORCE_FIELD_FRAGMENT_DIM - 1;
+  Vec3 closest = playerCoord.clamp(fragmentGridCoord, fragmentGridCoord + vec3(range, range, range));
+
+  requireInPlayerInfluence(playerCoord, closest);
+  return closest;
+}
+
+// Helper function to clamp a value between min and max
+function clamp(int32 value, int32 min, int32 max) pure returns (int32) {
```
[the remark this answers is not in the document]
vdrg: whoops was using it before, ty

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -63,6 +63,27 @@ function requireInPlayerInfluence(Vec3 playerCoord, EntityId entityId) view retu
   return coord;
 }
 
+// Checks if the player is in range of the fragment (8x8x8 cube)
+function requireFragmentInPlayerInfluence(Vec3 playerCoord, EntityId fragmentEntityId) view returns (Vec3) {
```
dhvanipa, line 67: this is neat, what I did in the past is make the client pass in the force field coord, but this lets us skip that

packages/world/src/ObjectTypes.sol
```diff
@@ -224,6 +224,7 @@ library ObjectTypes {
   // Misc
   // ------------------------------------------------------------
   ObjectTypeId constant Player = ObjectTypeId.wrap(Misc | 0);
+  ObjectTypeId constant ForceFieldFragment = ObjectTypeId.wrap(Misc | 1);
```
alvrs, line 227: Added this to the `objects.ts` file, from which this file is generated: https://github.com/tenetxyz/biomes-client/commit/afd77bc6da9af9a89f2a154421d98fb6964641ac

PR #86, opened 2025-03-06
Add fall damage threshold

dhvanipa:
Fixes https://github.com/tenetxyz/biomes-contracts/issues/67


packages/world/test/BiomesAssertions.sol
```diff
@@ -88,8 +88,8 @@ abstract contract BiomesAssertions is MudTest, GasReporter {
   function assertEnergyFlowedFromPlayerToLocalPool(
```
vdrg, line 88: would it be cleaner to instead add the expected loss as an argument instead? and assertLt() inside?

packages/world/test/Gravity.t.sol
```diff
@@ -99,7 +103,14 @@ contract GravityTest is BiomesTest {
     assertTrue(ObjectType.get(mineEntityId) == ObjectTypes.Air, "Mine entity is not air");
     assertInventoryHasObject(aliceEntityId, mineObjectTypeId, 1);
     EnergyDataSnapshot memory afterEnergyDataSnapshot = getEnergyDataSnapshot(aliceEntityId, playerCoord);
-    assertEnergyFlowedFromPlayerToLocalPool(beforeEnergyDataSnapshot, afterEnergyDataSnapshot);
+    uint128 playerEnergyLost = assertEnergyFlowedFromPlayerToLocalPool(
+      beforeEnergyDataSnapshot,
+      afterEnergyDataSnapshot
+    );
+    assertTrue(
```
vdrg, line 110: There's assertLt

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -61,10 +62,14 @@ library MoveLib {
       if (gravityAppliesForMove) {
         if (oldBaseCoord.y() < newBaseCoord.y()) {
           numJumps++;
+          currentFallHeight = 0;
```
vdrg, line 65: I think we might not need this new variable, can't we just track the total falls and then take the min(numFalls - threshold, 0)?

vdrg (approved):
lgtm but some improvements are possible

packages/world/test/Gravity.t.sol
dhvanipa: TIL `assertLt` and `assertGt`

packages/world/test/BiomesAssertions.sol
dhvanipa: but then for each test we'd have to calculate the expected loss, right now I just want _some_ loss, dont care about the amount

packages/world/test/Gravity.t.sol
dhvanipa: updated ALL

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: but we want consecutive falls, if we just track `numFalls` like before, we won't know if they fell one by one which would result in no damage, or 4 in one go

vdrg: ah ok

packages/world/test/BiomesAssertions.sol
vdrg: what about defining an overloaded function with the same name? That way you don't have to duplicate the check everywhere

dhvanipa: > That way you don't have to duplicate the check everywhere
wdym?

vdrg: have an additional assertEnergyFlowedFromPlayerToLocalPool that receives an extra parameter and does the require internally

dhvanipa: ahh okay will consider in future PR

PR #88, opened 2025-03-06
Add admin system

dhvanipa:
- This is so we can give items to players when testing locally


packages/world/test/utils/TestUtils.sol
```diff
@@ -55,23 +57,28 @@ library TestUtils {
...
+    return _addToolToInventory(ownerEntityId, toolObjectTypeId);
+  }
+
+  function removeFromInventory(
     EntityId ownerEntityId,
     ObjectTypeId objectTypeId,
     uint16 numObjectsToRemove
   ) public asWorld {
+    require(!objectTypeId.isTool(), "To remove a tool, you must pass in the tool entity id");
     _removeFromInventoryCount(ownerEntityId, objectTypeId, numObjectsToRemove);
```
vdrg, line 75: should we rename the original functions too?

vdrg (approved):

vdrg:
also maybe this could be a test system that is only deployed on tests, so we don't have to deploy it to prod

dhvanipa:
> also maybe this could be a test system that is only deployed on tests, so we don't have to deploy it to prod

We need it outside of the tests, ie when testing in the client. I think its fine to deploy to prod since it's checking namespace owner and as long as the owner exists they can always write to the tables bypassing the systems. When the namespace owner is removed, this system would also stop working

packages/world/test/utils/TestUtils.sol
dhvanipa: yee renamed!

PR #97, opened 2025-03-10
Fix player existence check in spawn system

### LINKED ISSUE #95: Spawn system currently checks if player entity exists
As now we determine if a player is dead by checking that energy == 0, we should instead check this in the spawn system.

vdrg:
Fixes #95


packages/world/src/systems/SpawnSystem.sol
```diff
@@ -84,20 +84,36 @@ contract SpawnSystem is System {
   function isValidSpawn(Vec3 spawnCoord) public view returns (bool) {
     Vec3 belowCoord = spawnCoord - vec3(0, 1, 0);
     Vec3 topCoord = spawnCoord + vec3(0, 1, 0);
+
     ObjectTypeId spawnObjectTypeId = getObjectTypeIdAt(spawnCoord);
+    if (
+      spawnObjectTypeId.isNull() ||
```
dhvanipa, line 90: cleannnn

dhvanipa (approved):

PR #100, opened 2025-03-11
Remove world border

### LINKED ISSUE #91: Do we need world border checks if the borders are restricted by what chunks are explored?
Since we provide the chunks, there is already an implicit border, so maybe we dont need an explicit one in the contracts then

vdrg:
Closes #91


packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -147,10 +145,6 @@ library MoveLib {
 
   function _gravityApplies(Vec3 playerCoord) internal view returns (bool) {
     Vec3 belowCoord = playerCoord - vec3(0, 1, 0);
-    // We don't want players to fall off the edge of the world
-    if (!inWorldBorder(belowCoord)) {
```
dhvanipa, line 151: What happens now if you're at the bottom of the world and mine it?

Does it not let you mine it? or can you stand on "void"?

dhvanipa: well bottom of the last explored chunk**

vdrg: hmm I believe it won't let you mine it, because it will try to apply gravity and it will encounter a chunk not explored yet

dhvanipa: gotcha so if you're standing above the world border, ppl cant mine it and make you fall - makes sense
you can mine blocks just above the world border, and there'll be a "void" -- makes sense
you cant "move" or fall into the void as the txs wil fail -- makes sense
even if you make a little island and break all the blocks, ppl can still build to get to you - makes sense

dhvanipa (approved):

Issue #108, opened 2025-03-11
Seeds


vdrg:
- New object for seeds
- Allow planting seeds in wet tilled soil
  - Remove energy from local pool and transfer to the sapling
  - When enough time has passed, mining the placed seed will return wheat (or something else depending on the seed)

vdrg:
Implemented in #111

PR #111, opened 2025-03-12
Farming

### LINKED ISSUE #105: Support getting a different object from mining a block
- Tall grass block can give you seeds
- Mining crops before they are fully grown gives you nothing

vdrg:
- Adds wheat seeds
- Tracks build time for seeds
- Different drops depending on if the seed had enough time to grow

TODO: fix some tests

Closes #105, #108


packages/world/src/systems/MineSystem.sol
```diff
@@ -170,6 +172,31 @@ contract MineSystem is System {
...
 
+  function _handleDrop(
+    EntityId playerEntityId,
+    EntityId baseEntityId,
+    Vec3 mineCoord,
+    ObjectTypeId mineObjectTypeId
+  ) internal {
+    if (mineObjectTypeId == ObjectTypes.FescueGrass) {
+      // TODO: add randomness?
+      addToInventory(playerEntityId, ObjectTypes.Player, ObjectTypes.WheatSeeds, 1);
```
vdrg, line 183: add todo: conservation of fescue grass

packages/world/src/systems/init/InitTerrainBlocksSystem.sol
```diff
@@ -138,5 +141,16 @@ contract InitTerrainBlocksSystem is System {
     createTerrainBlock(ObjectTypes.GoldOre, 100);
     createTerrainBlock(ObjectTypes.DiamondOre, 100);
     createTerrainBlock(ObjectTypes.NeptuniumOre, 100);
+
+    // TODO: move to a different init system?
+    // TODO: Should these have inventory slots?
+    // TODO: is this mass calculation correct?
+    uint128 wheatEnergy = 1000;
+    createPassableTerrainBlock(ObjectTypes.Wheat, uint32(100 + energyToMass(wheatEnergy)));
```
vdrg, line 149: should be 0 mass, only energy

packages/world/src/systems/MineSystem.sol
```diff
@@ -158,7 +159,11 @@ library MineLib {
 contract MineSystem is System {
   using ObjectTypeLib for ObjectTypeId;
 
-  function _removeBlock(EntityId entityId, Vec3 coord) internal {
+  function _removeBlock(EntityId entityId, ObjectTypeId objectTypeId, Vec3 coord) internal {
+    if (objectTypeId.isSeed()) {
```
dhvanipa, line 163: What happens if you call mine on a seed that is fully grown? You'd get the seed right. That feels odd cuz the seed has already grown

Feel like we should either revert the mine, and player has to call growSeed, or we should call growSeed here then run mine

packages/world/src/systems/FarmingSystem.sol
```diff
@@ -34,4 +35,24 @@ contract FarmingSystem is System {
 
     ObjectType._set(farmlandEntityId, ObjectTypes.Farmland);
   }
+
+  function growSeed(Vec3 coord) external {
+    // TODO: should this function be open?
```
dhvanipa, line 40: Yee I think it should be open, this is similar to how nature system will be, it will be open and this is kind of the first function in the nature system, could even put it in a new system

vdrg: but do we still need proximity checks? And if we ever add incentives we would need a valid player right?

packages/world/src/systems/MineSystem.sol
vdrg: hmm but shouldn't the client figure that out? could even do a batch call, that way we keep mining cheaper and simpler

dhvanipa: our client will deff do it (but many ppl will run scripts), and so I mean the way it is right now, technically the physics allows you to store energy in a seed and then put it back into the local energy pool, by mining seeds instead of calling grow. that feels weird and shouldnt be allowed

packages/world/src/systems/FarmingSystem.sol
dhvanipa: I dont think we need proximity checks for stuff like this or really any nature system calls. it's just progressing the world by calling tick. Yea we could require a player so we can give incentives in the future

packages/world/src/systems/MineSystem.sol
vdrg: how would this work in the case of trees? should it mine the base of the trunk or something like that? maybe we can just prevent it for now

dhvanipa: yea simplest might be to just revert the mine tx

dhvanipa:
also if you mine `WetFarmland` you should get dirt right? I think right now it gives you the wet farmland but then you could place it again

packages/world/src/systems/MineSystem.sol
vdrg: I guess this also applies for seeds that are on top of blocks being mined

dhvanipa: yee just revert. Our client will never get into this state anyways so user wont know about it

packages/world/src/ObjectTypeLib.sol
```diff
@@ -114,6 +114,66 @@ library ObjectTypeLib {
...
+
+    amounts = new ObjectAmount[](1);
+    amounts[0] = ObjectAmount(objectTypeId, 1);
+    return amounts;
+  }
+
+  function timeToGrow(ObjectTypeId objectTypeId) internal pure returns (uint128) {
+    // TODO: different times for different seeds
+    if (objectTypeId.isSeed()) {
+      return 15 minutes;
```
frolic, line 171: TIL you can use `minutes` 😮

packages/world/mud.config.ts
```diff
@@ -378,6 +378,16 @@ export default defineWorld({
       key: [],
     },
     // ------------------------------------------------------------
+    // Farming
+    // ------------------------------------------------------------
+    SeedGrowth: {
+      schema: {
+        entityId: "EntityId",
+        fullyGrownAt: "uint128",
```
frolic, line 386: not sure what our patterns are here so feel free to ignore, but I found in some projects I was fiddling with that it was a bit more flexible to use something like `plantedAt` and then adjust the time-to-grow instead of combining the time-to-grow with the planted-at timestamp

vdrg: I don't think we have spent too much effort yet on trying to follow a consistent set patterns, I was hoping that after we finish the main features we could do a pass where we analyze the schemas and try to refactor as much as we can for efficiency and consistency.

Having said that, I would love to better understand the benefits in flexibility you mention, as originally I had used plantedAt and timeToGrow for each seed, but ended up with just fullyGrownAt to make it a bit cheaper (and if we need to get planted at we can just do fullyGrownAt - timeToGrow).

frolic: fullyGrownAt - timeToGrow wouldn't be accurate if you adjust the timeToGrow constant after setting fullyGrownAt in storage

vdrg: Yeah agreed, maybe the optimization is not worth it, however if we in the future wanted to adjust timeToGrow we would need to update for all seeds, the idea was to be able to have something like fertilizer so that you decrease the time to grow. With fullyGrownAt you can still do this if you assume the initial timeToGrow is constant.

frolic: ahh, modifiers makes sense

in the case of fertilizer, would you be decreasing the `fullyGrownAt`? and would you still have to track if fertilizer was used or not to avoid multiple uses? or maybe each use is a multiplier (e.g. `* 0.5`) rather than subtracting a constant, so it has diminishing returns and rarely reach zero with just fertilizer

vdrg: yeah something like that. Another option would be to store both plantedAt + timeToGrow, and just adjust timeToGrow when using fertilizer. I guess it's fine if we adjust the initial timeToGrow constant without updating all the existing trees, and just apply them for future ones.

PR #113, opened 2025-03-13
feat: explore region energy

alvrs:
corresponding client PR: https://github.com/tenetxyz/biomes-client/pull/547


packages/world/src/Constants.sol
```diff
@@ -13,8 +10,9 @@ uint16 constant PLAYER_FALL_DAMAGE_THRESHOLD = 3;
 
 uint256 constant SPAWN_BLOCK_RANGE = 10;
 
+int32 constant FRAGMENT_SIZE = 8;
 int32 constant CHUNK_SIZE = 16;
-int32 constant AREA_SIZE = 512;
+int32 constant REGION_SIZE = 512;
```
vdrg, line 15: maybe add comments about fragments and chunks being 3d vs regions being 2d

vdrg (approved):
lgtm. The reason for not making regions the same size as chunks and encoding in the chunk data is because shunks are 3d while regions are 2d, correct? And the client will set the region energy when exploring the first chunk of that region?

alvrs:
> The reason for not making regions the same size as chunks

Main reason is that we want the energy pool to be shared for a larger area than individual pools per chunk. But yes also want it to be 2d instead of 3d.

alvrs:
> And the client will set the region energy when exploring the first chunk of that region?

Correct, client will explore the region energy when it's close enough (in an adjacent chunk)

dhvanipa:
We can remove `PHYSICS_ORACLE_ADDRESS` since it was only being used in the `OracleSystem`

PR #114, opened 2025-03-13
Only store surface chunks for random spawn

vdrg:


dhvanipa:
We're only using `ExploredChunk` for random spawn? So this change is to save gas right? Can we no read the isSurface from the bytecode? Or am I confusing two different things lol

vdrg:
> We're only using `ExploredChunk` for random spawn? So this change is to save gas right? Can we no read the isSurface from the bytecode? Or am I confusing two different things lol

we need to spawn only on surface chunks, so we need to choose only from surface chunks. If we stored all of the chunks it would be hard to get a surface chunk from the randomness provided by the previous 10 hashes

dhvanipa (approved):

PR #115, opened 2025-03-13
feat: trees

vdrg:


packages/world/src/systems/MineSystem.sol
```diff
@@ -234,21 +230,25 @@ contract MineSystem is System {
...
             _handleDrop(playerEntityId, aboveTypeId);
+            addEnergyToLocalPool(coord, ObjectTypeMetadata._getEnergy(aboveTypeId));
           }
         }
 
-        _removeBlock(baseEntityId, mineObjectTypeId, baseCoord);
+        _removeBlock(baseEntityId, baseCoord);
         _handleDrop(playerEntityId, mineObjectTypeId);
 
+        if (mineObjectTypeId.isSeed()) {
```
dhvanipa, line 242: ahhh I guess this works cuz seeds cant be on top of other seeds

packages/world/src/utils/MathUtils.sol
```diff
@@ -18,6 +18,15 @@ function mod(int256 x, int256 y) pure returns (uint256) {
   return uint256(((x % y) + y) % y);
 }
 
+// TODO: we should import math utils from solady or OZ
+function min(int256 x, int256 y) pure returns (int256) {
```
dhvanipa, line 22: I assume can resolve this once we merge in main

packages/world/src/ObjectTypes.sol
```diff
@@ -186,7 +186,16 @@ library ObjectTypes {
   ObjectTypeId constant DiamondOre = ObjectTypeId.wrap(Block | 270);
   ObjectTypeId constant NeptuniumOre = ObjectTypeId.wrap(Block | 271);
   ObjectTypeId constant WetFarmland = ObjectTypeId.wrap(Block | 272);
+  // TODO: rename WheatSeeds to WheatSeed (client)
```
dhvanipa, line 189: I can update this before we merge it in

packages/world/script/initObjects.sol
```diff
@@ -539,6 +539,15 @@ function initObjects() {
     ObjectTypes.WheatSeeds,
     ObjectTypeMetadataData({ stackable: 99, maxInventorySlots: 0, mass: 0, energy: 100, canPassThrough: true })
   );
+  // TODO: tree seeds must match the total energy of its components
```
dhvanipa, line 542: I assume we'll pick some energy value based on the average tree. Then we return the remaining to energy pool anyways

vdrg: Yeah exactly, as long as it is higher or equal to the maximum possible energy for the tree this will work

dhvanipa: with the randomness stuff we'll know the max possible tree size right?

vdrg: The maximum size won't depend on randomness, I can write a formula later given the tree params

dhvanipa (approved):

PR #117, opened 2025-03-14
feat: add region seeding + merkle proof verification

alvrs:
corresponding client PR: https://github.com/tenetxyz/biomes-client/pull/556


packages/world/script/InitTerrain.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+/* Autogenerated file. Do not edit manually. */
+
+import { RegionMerkleRoot } from "../src/codegen/tables/RegionMerkleRoot.sol";
+
+contract InitTerrain {
```
vdrg, line 8: these don't need to be contracts right? they can just be file level functions

packages/world/src/systems/TerrainSystem.sol
```diff
@@ -4,16 +4,25 @@ pragma solidity >=0.8.24;
 import { System } from "@latticexyz/world/src/System.sol";
 import { Vec3 } from "../Vec3.sol";
 import { ExploredChunkCount } from "../codegen/tables/ExploredChunkCount.sol";
+import { RegionMerkleRoot } from "../codegen/tables/RegionMerkleRoot.sol";
 import { ExploredChunkByIndex, ExploredChunk, InitialEnergyPool, LocalEnergyPool } from "../utils/Vec3Storage.sol";
 import { SSTORE2 } from "../utils/SSTORE2.sol";
-import { INITIAL_ENERGY_PER_VEGETATION } from "../Constants.sol";
+import { INITIAL_ENERGY_PER_VEGETATION, REGION_SIZE, CHUNK_SIZE } from "../Constants.sol";
+import { MerkleProof } from "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
```
vdrg, line 11: solady also has a merkle proof library, which is probably more efficient (I'm sure oz is fine too)

packages/world/src/systems/libraries/TerrainLib.sol
```diff
@@ -138,4 +138,15 @@ library TerrainLib {
...
   }
+
+  /// @dev Get the leaf hash of a chunk
+  function _getChunkLeafHash(Vec3 chunkCoord, bytes memory chunk) internal pure returns (bytes32) {
+    return keccak256(bytes.concat(keccak256(abi.encode(chunkCoord, chunk))));
+  }
+
+  /// @dev Get the vegetation leaf hash
+  function _getVegetationLeafHash(uint32 vegetationCount) internal pure returns (bytes32) {
+    bytes32 key = keccak256(abi.encode("region.vegetation"));
```
vdrg, line 149: maybe define as constant?

vdrg (approved):
nice!

packages/world/mud.config.ts
```diff
@@ -205,6 +205,14 @@ export default defineWorld({
       },
       key: [],
     },
+    RegionMerkleRoot: {
```
dhvanipa, line 208: Should we make this table 3d? ie have a y value, would align with the the energy pool table

dhvanipa: yea fair, then lets make energy region table 2D too!

packages/world/src/systems/TerrainSystem.sol
alvrs: would leave for a followup if you agree!

packages/world/mud.config.ts
dhvanipa: yea sure!

PR #118, opened 2025-03-15
Fix mining bug

dhvanipa:
- The amount of mass the tool loses should be 1-1 with how much mass is taken from the block + base mass from energy spent. Previously it was taking more mass from the tool than the block


packages/world/src/systems/MineSystem.sol
```diff
@@ -231,9 +235,10 @@ contract MineSystem is System {
         {
           // Remove seeds placed on top of this block
           Vec3 aboveCoord = baseCoord + vec3(0, 1, 0);
-          (EntityId aboveEntityId, ObjectTypeId aboveTypeId) = getOrCreateEntityAt(aboveCoord);
+          ObjectTypeId aboveTypeId = safeGetObjectTypeIdAt(aboveCoord);
```
dhvanipa, line 238: Did this so we don't unnecessarily create entities mining stuff

packages/world/src/systems/MineSystem.sol
```diff
@@ -108,10 +108,14 @@ library MineLib {
   }
 
   function _processMassReduction(EntityId playerEntityId, EntityId minedEntityId) public returns (uint128) {
-    (uint128 toolMassReduction, ) = useEquipped(playerEntityId);
-    uint128 totalMassReduction = energyToMass(PLAYER_MINE_ENERGY_COST) + toolMassReduction;
     uint128 massLeft = Mass._getMass(minedEntityId);
-    return massLeft <= totalMassReduction ? 0 : massLeft - totalMassReduction;
+    uint128 baseMassReduction = energyToMass(PLAYER_MINE_ENERGY_COST);
+    if (massLeft <= baseMassReduction) {
```
vdrg, line 113: just trying to understand this, if massLeft is <= baseMassReduction then `useEquipped` is not called, so the tool would not be removed?

dhvanipa: yee, if the player can mine the block's remaining mass with just the energy they are spending, then we dont use the tool

packages/world/src/systems/MineSystem.sol
```diff
@@ -231,9 +235,10 @@ contract MineSystem is System {
         {
           // Remove seeds placed on top of this block
           Vec3 aboveCoord = baseCoord + vec3(0, 1, 0);
-          (EntityId aboveEntityId, ObjectTypeId aboveTypeId) = getOrCreateEntityAt(aboveCoord);
+          ObjectTypeId aboveTypeId = safeGetObjectTypeIdAt(aboveCoord);
           if (aboveTypeId.isSeed()) {
             _requireSeedNotFullyGrown(baseEntityId);
+            EntityId aboveEntityId = createEntityAt(aboveCoord, aboveTypeId);
```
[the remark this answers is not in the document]
dhvanipa: ahh yea its incorrect, fixed

packages/world/src/systems/MineSystem.sol
vdrg: wait but shouldn't we also change the amount of energy transferred to the pool then? lmk if I understand this correctly:
- mine energy cost is removed from the player
- mass is removed from the tool
- mass is removed from the object being mined

shouldn't the amount of energy added to the pool be the sum of those 3 quantities?

dhvanipa: lets go over this in tuning, added a todo

PR #120, opened 2025-03-17
Replace math utils and randomness with solady

vdrg:


dhvanipa (approved):
ooc did this increase/decrease the gas?

vdrg:
checking now, also replaced the way we generated pseudo random numbers @dhvanipa

vdrg:
yep gas savings across the board

packages/world/src/systems/SpawnSystem.sol
```diff
@@ -117,7 +115,7 @@ contract SpawnSystem is System {
   }
 
   function getValidSpawnY(Vec3 spawnCoordCandidate) public view returns (Vec3 spawnCoord) {
-    for (int32 i = 0; i < CHUNK_SIZE; i++) {
+    for (int32 i = CHUNK_SIZE - 1; i >= 0; i--) {
```
dhvanipa, line 118: ahh nice, iterating from top to bottom cuz likely surface is on top so we can early return

dhvanipa:
> checking now, also replaced the way we generated pseudo random numbers @dhvanipa

should we use the same for other places like ores?

vdrg:
done for ores!

PR #121, opened 2025-03-17
FOOD

vdrg:


packages/world/script/initObjects.sol
```diff
@@ -417,7 +417,7 @@ function initObjects() {
   );
   ObjectTypeMetadata.set(
     ObjectTypes.Wheat,
-    ObjectTypeMetadataData({ stackable: 99, maxInventorySlots: 0, mass: 1, energy: 0, canPassThrough: true })
+    ObjectTypeMetadataData({ stackable: 99, maxInventorySlots: 0, mass: 0, energy: 100, canPassThrough: true })
```
dhvanipa, line 420: nice, will update this in client but can merge it in since this is the only food

PR #122, opened 2025-03-17
fix: Make ore type fixed after partial mine

### LINKED ISSUE #77: On mine of ores, not taking into account ore mass

vdrg:
- After the first mine, ore type becomes fixed
- Ore mass is now set correctly

Fixes #77


dhvanipa (approved):
lgtm!

PR #123, opened 2025-03-17
refactor: Remove withExtraData functions

### LINKED ISSUE #62: Remove `WithExtraData` from function names and instead use overloaded functions
Would be cleaner if we could name the functions all the same

vdrg:
gas increase seems to be around 100-200 gas

Fixes #62


packages/world/src/systems/BuildSystem.sol
```diff
@@ -96,11 +96,11 @@ library BuildLib {
 contract BuildSystem is System {
   using ObjectTypeLib for ObjectTypeId;
 
-  function buildWithExtraData(
+  function buildWithDirection(
     ObjectTypeId buildObjectTypeId,
     Vec3 baseCoord,
     Direction direction,
-    bytes memory extraData
+    bytes calldata extraData
```
dhvanipa, line 103: ahh nice I assume using calldata over memory also saves gas?

dhvanipa (approved):

packages/world/src/systems/BuildSystem.sol
vdrg: Yeah in some cases, when the function doesn't need to copy it to memory

PR #126, opened 2025-03-18
feat: kill players 🔫 

### LINKED ISSUE #84: when energy reaches 0 on any action, player should die

vdrg:
Fixes #84


packages/world/mud.config.ts
```diff
@@ -330,10 +322,11 @@ export default defineWorld({
       },
       key: ["entityId"],
     },
-    ForceField: {
+    Machine: {
       schema: {
         entityId: "EntityId",
         createdAt: "uint128",
+        accDepletedTime: "uint128",
```
frolic, line 329: what's "acc" here?

vdrg: accumulated, can change to that if you prefer

frolic: generally find these abbreviations hard to follow, so would prefer the whole word

but would `depletedTime` not be descriptive enough? or `totalDepletedTime`?

vdrg: accumulated made sense to me because it accumulates the different periods where it was depleted, `depletedTime` sounds like "time since it got depleted" imo. `total` works I guess

frolic: we seem to use `At` suffix for timestamps so `Time` or `Seconds` or something would make sense to me, as long as we're consistent with it

vdrg: Sounds good, yeah we need to do a pass to improve consistency for sure

packages/world/src/systems/BedSystem.sol
```diff
@@ -32,18 +33,17 @@ import { EntityId } from "../EntityId.sol";
 
 // To avoid reaching bytecode size limit
 library BedLib {
-  function transferInventory(EntityId playerEntityId, EntityId bedEntityId, ObjectTypeId objectTypeId) public {
-    transferAllInventoryEntities(playerEntityId, bedEntityId, objectTypeId);
-  }
+  function transferInventory(EntityId playerEntityId, EntityId bedEntityId, ObjectTypeId objectTypeId) public {}
```
dhvanipa, line 36: why is there an empty function here?

vdrg: Ah damn I missed that, we need to add a test for transferring inventory

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -30,110 +31,132 @@ import { MAX_PLAYER_INFLUENCE_HALF_WIDTH, PLAYER_ENERGY_DRAIN_RATE, FRAGMENT_SIZ
...
+    checkWorldStatus();
+    EntityId playerEntityId = Player._get(player);
+    require(playerEntityId.exists(), "Player does not exist");
+    require(!PlayerStatus._getBedEntityId(playerEntityId).exists(), "Player is sleeping");
+    Vec3 playerCoord = PlayerPosition._get(playerEntityId);
 
-  EnergyData memory playerEnergyData = updateEnergyLevel(playerEntityId);
-  require(playerEnergyData.energy > 0, "Player is dead");
+    EnergyData memory playerEnergyData = updatePlayerEnergy(playerEntityId);
+    require(playerEnergyData.energy > 0, "Player is dead");
```
dhvanipa, line 44: so the idea is if you're dead, then all txs like build/mine/move/drop etc will revert, and so you need to call spawn

in the client, we'll know if they're dead, so no need to call activate, just show death/respawn screen

we'll still use activate to kill players who never load the client

packages/world/src/systems/BedSystem.sol
dhvanipa: if you'd like to add in a separate PR could add a stub with a todo so we dont forget

dhvanipa (approved):

packages/world/src/utils/PlayerUtils.sol
dhvanipa: but also this means if you mine/build and lose energy, it doesnt kill you right?

dhvanipa: do you think its better if it kills you ie remove from grid? or better if client shows just shows death and only removes you once you call spawn/activate

vdrg: It kills you in the sense that it can leave you with 0 energy, but yeah you need to respawn or call activate. We could kill whenever energy gets to 0, but it was a little bit more complex

vdrg: > do you think its better if it kills you ie remove from grid? or better if client shows just shows death and only removes you once you call spawn/activate

Not sure tbh, if we are worried about consistency then maybe we can just kill whenever energy gets to 0

dhvanipa: I think its fine for build/mine but I'm worried about moves, and later hits. If you mine a block and drop someone, you'd expect them to be removed from the grid. Same with hitting. Otherwise the client would have to call `activate` on these players

vdrg: Ah we also need to take into account gas costs, having too many items could make you prevent certain deaths if we remove on death

dhvanipa: true, this will be true for hits too. In these cases though, I assume the attacker is willing to pay the gas cuz you want the other persons items. As long as the tx fits within block limit haha. This code path should also be rare

PR #131, opened 2025-03-19
feat: getRandomOreType

### LINKED ISSUE #130: Add view function to get random ore type
- It should only work if there is an existing commitment

vdrg:
Closes #130


alvrs:
works great

PR #144, opened 2025-03-19
add death notif

dhvanipa:
- also added timestamp field
- fixed bug in transferring to air on death, would transfer to zero entity


vdrg (approved):
Lgtm!

Issue #141, opened 2025-03-19
make players have 0 mass


dhvanipa:
dont take into player mass when spawning

PR #147, opened 2025-03-20
Energy improvements

### LINKED ISSUE #146: Make energy == mass
There's no reason to have a multiplier

vdrg:
- Remove player mass
- Remove energy<>mass conversion
- Remove energy cost for drop/pickup/transfer

Also included the change to make `pnpm dev` use `mud dev-contracts` so we don't have to redeploy on every change

Fixes #146, #141, #139


alvrs (approved):
lgtm

PR #148, opened 2025-03-20
Kill player on decrease energy

vdrg:
- We use a new decreasePlayerEnergy util that kills the player if energy reaches 0
- Included the fix to first decrease mass from tool before consuming player energy

### DIFF
--- a/packages/world/src/systems/MineSystem.sol
+++ b/packages/world/src/systems/MineSystem.sol
@@ -25,9 +25,9 @@ import { OreCommitment } from "../utils/Vec3Storage.sol";
 import { getUniqueEntity } from "../Utils.sol";
 import { addToInventory, useEquipped } from "../utils/InventoryUtils.sol";
 import { PlayerUtils } from "../utils/PlayerUtils.sol";
-import { updateMachineEnergy, transferEnergyToPool, addEnergyToLocalPool, updateSleepingPlayerEnergy } from "../utils/EnergyUtils.sol";
+import { updateMachineEnergy, updatePlayerEnergy, addEnergyToLocalPool, updateSleepingPlayerEnergy, decreasePlayerEnergy } from "../utils/EnergyUtils.sol";
 import { getForceField, destroyForceField } from "../utils/ForceFieldUtils.sol";
-import { notify, MineNotifData } from "../utils/NotifUtils.sol";
+import { notify, MineNotifData, DeathNotifData } from "../utils/NotifUtils.sol";
 import { getOrCreateEntityAt, getObjectTypeIdAt, createEntityAt, getEntityAt, getPlayer } from "../utils/EntityUtils.sol";
 import { callProgramOrRevert } from "../utils/callProgram.sol";
 
@@ -71,7 +71,7 @@ contract MineSystem is System {
   }
 
   function getRandomOreType(Vec3 coord) external view returns (ObjectTypeId) {
-    (ObjectTypeId ore, ) = MineLib._getRandomOreType(coord);
+    (ObjectTypeId ore, ) = RandomOreLib._getRandomOreType(coord);
     return ore;
   }
 
@@ -81,8 +81,6 @@ contract MineSystem is System {
 
     (EntityId entityId, ObjectTypeId mineObjectTypeId) = getOrCreateEntityAt(coord);
     require(mineObjectTypeId.isMineable(), "Object is not mineable");
-
-    transferEnergyToPool(playerEntityId, playerCoord, PLAYER_MINE_ENERGY_COST);
 
     EntityId baseEntityId = entityId.baseEntityId();
     Vec3 baseCoord = Position._get(baseEntityId);
@@ -95,7 +93,7 @@ contract MineSystem is System {
     } else if (mineObjectTypeId.isSeed()) {
       _requireSeedNotFullyGrown(baseEntityId);
     } else if (mineObjectTypeId == ObjectTypes.AnyOre) {
-      mineObjectTypeId = MineLib._mineRandomOre(baseEntityId, coord);
+      mineObjectTypeId = RandomOreLib._mineRandomOre(baseEntityId, coord);
     }
 
     // First coord will be the base coord, the rest is relative schema coords
@@ -103,7 +101,7 @@ contract MineSystem is System {
 
     uint128 finalMass;
     {
-      finalMass = MineLib._processMassReduction(playerEntityId, baseEntityId);
+      finalMass = MassReductionLib._processMassReduction(playerEntityId, baseEntityId, playerCoord);
       if (finalMass == 0) {
         if (mineObjectTypeId == ObjectTypes.Bed) {
           // If mining a bed with a sleeping player, kill the player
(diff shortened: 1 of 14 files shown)


packages/world/src/systems/MineSystem.sol
```diff
@@ -176,6 +174,83 @@ contract MineSystem is System {
...
+    uint128 playerEnergyReduction = 0;
+
+    // if tool mass reduction is not enough, consume energy from player up to mine energy cost
+    if (toolMassReduction < massLeft) {
+      uint128 remaining = massLeft - toolMassReduction;
+      playerEnergyReduction = PLAYER_MINE_ENERGY_COST <= remaining ? PLAYER_MINE_ENERGY_COST : remaining;
+      decreasePlayerEnergy(playerEntityId, playerCoord, playerEnergyReduction);
+    }
+
+    uint128 totalEnergyReduction = playerEnergyReduction + toolMassReduction;
```
dhvanipa, line 246: why is the mass being added to the local pool? the mass is mined but its in the inventory and will appear back when its built. So should not be converted into energy, otherwise total energy in the world will increase

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -118,11 +118,16 @@ library MoveLib {
...
 
+    uint128 currentEnergy = Energy._getEnergy(playerEntityId);
     uint128 energyCost = (PLAYER_MOVE_ENERGY_COST * uint128(newBaseCoords.length - numFalls)) +
       (PLAYER_FALL_ENERGY_COST * numFalls);
-    uint128 currentEnergy = Energy._getEnergy(playerEntityId);
-    transferEnergyToPool(playerEntityId, playerCoord, energyCost > currentEnergy ? currentEnergy : energyCost);
-    // TODO: drop inventory items
+    energyCost = energyCost > currentEnergy ? currentEnergy : energyCost;
+
+    // TODO: figure out how to run gravity and only kill AFTER that
```
dhvanipa, line 126: so now if the player dies while falling what happens? does gravity keep running or the transaction ends?

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -135,6 +135,7 @@ function useEquipped(
     } else {
       // use 10% of the mass if it's greater than 10
       if (toolMassLeft > 10) {
+        // TODO: shouldn't it be a percentage of the max mass?
```
dhvanipa, line 138: but you might not be able to use 10% of the max if you have less than that right?

packages/world/src/utils/EnergyUtils.sol
```diff
@@ -76,17 +77,42 @@ function updatePlayerEnergy(EntityId playerEntityId) returns (EnergyData memory)
 
   if (energyDrained > 0) {
     addEnergyToLocalPool(coord, energyDrained);
-  }
 
-  // Player is dead
-  if (energyData.energy == 0) {
-    PlayerUtils.killPlayer(playerEntityId, coord);
+    if (energyData.energy == 0) {
```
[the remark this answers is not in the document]
vdrg: Exactly

packages/world/src/systems/libraries/MoveLib.sol
vdrg: That's what we need t figure out, we don't want players to die in the middle of a fall right, so we should probably wait until they hit the ground

packages/world/src/utils/InventoryUtils.sol
vdrg: In that case you would use all the remaining mass maybe. Idk, it didn't make sense to me conceptually that every time you mine it uses less mass

packages/world/src/systems/MineSystem.sol
vdrg: Ah you mean when it is burned?

packages/world/src/utils/InventoryUtils.sol
dhvanipa: yea agree what you're suggesting makes more sense. lets change it to that! could you create an issue for it?

packages/world/src/systems/MineSystem.sol
vdrg: Nvm I think I get it

vdrg: Nvm I think I get it

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: oh no we shouldn't wait for them to hit the ground! they should die in the middle of a fall

otherwise imagine you're standing on a mountain/floating island, and jump off. You'd have to pass gas to run gravity alll the way to the ground. vs the fall damage parameter is basically a way we can tune the max gas you have to pay to die, cuz we know for sure gravity will stop running after X blocks.

vdrg: ah interesting, but then the inventory would stay floating in the air

vdrg: make sense, would it become cheaper if we do that thing where the client provides the whole path (including gravity) ?

packages/world/src/systems/MineSystem.sol
vdrg: wait I think I still don't get it lol the mass that gets removed from the tool has to go somewhere right? I understand not transferring the mass of the object being mined as it will eventually be placed / consumed, but the tool's mass has to go somewhere

dhvanipa: the tools mass will be returned to the world when it breaks --> ores used in crafting will respawn

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: we'd still have to verify the path in the contracts right? so wouldnt it be more expensive cuz compute is cheaper than call data?

vdrg: yeah maybe, haven't checked the move system in detail, but nvm we can look into it later

dhvanipa: can also make this todo into a separate github issue of dealing with gravity. I think when we fix it we should resolve the two missing tests in Gravity.t.sol as well that check for death

packages/world/src/systems/MineSystem.sol
vdrg: done

packages/world/src/systems/MineSystem.sol
```diff
@@ -176,6 +174,80 @@ contract MineSystem is System {
...
+  }
+}
+
+library MassReductionLib {
+  function _processMassReduction(
+    EntityId playerEntityId,
+    EntityId minedEntityId,
+    Vec3 playerCoord
+  ) public returns (uint128) {
+    // TODO: balancing, what should the proper mass and energy cost be?
```
dhvanipa, line 229: nit: can remove this todo now

packages/world/src/systems/libraries/MoveLib.sol
vdrg: I now made it so we just return that gravity doesn't apply if the energy was consumed while falling, so it will just kill the player mid air, will add tests for that after merge

PR #152, opened 2025-03-21
Refactor move lib

vdrg:
- Combine valid path checks with movement and cost calculation
- Optimize gravity calculations by extracting the logic to get the resulting coord after applying gravity (looping until a coord where gravity doesn't apply is found)
- Transfer energy to the pool at the final coordinate of the path
- If the player runs out of energy in the middle of the path, we stop moving and apply gravity, so even if the player dies mid air from energy depletion they are still moved to the ground, and their inventory is transferred to the final coordinate (no floating inventories)


packages/world/src/Vec3.sol
```diff
@@ -86,17 +86,17 @@ library Vec3Lib {
 
   function x(Vec3 a) internal pure returns (int32) {
     // Extract z component (leftmost 32 bits)
-    return int32(uint32((Vec3.unwrap(a) >> 64) & 0xFFFFFFFF));
```
dhvanipa, line 89: why was this removed? or added in the first place if its not needed?

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -151,28 +207,42 @@ library MoveLib {
     }
   }
 
-  function _gravityApplies(Vec3 playerCoord) internal view returns (bool) {
-    Vec3 belowCoord = playerCoord - vec3(0, 1, 0);
+  function runGravity(EntityId playerEntityId, Vec3 playerCoord) public {
```
dhvanipa, line 210: Compared to before, now with stacked players, the bottom player will first fall all the way, and then the player above will fall?

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -151,28 +207,42 @@ library MoveLib {
...
+    // Remove the current player from the grid
+    for (uint256 i = 0; i < playerCoords.length; i++) {
+      ReversePlayerPosition._deleteRecord(playerCoords[i]);
     }
-    if (getPlayer(belowCoord).exists()) {
-      return false;
+
+    (Vec3 finalCoord, uint128 totalCost) = _computeGravityResult(playerCoord, 0);
+
+    uint128 currentEnergy = Energy._getEnergy(playerEntityId);
```
dhvanipa, line 225: We need to read the energy by calling `updatePlayerEnergy` right? Otherwise if you mine a block below a player/stacked player case, their energy will be the one thats not updated

packages/world/src/Vec3.sol
vdrg: yeah the mask was not needed, as we are already casting it

packages/world/src/systems/libraries/MoveLib.sol
vdrg: yes sir, but I think it kind of result in the same end result?

vdrg: well except if it is a long fall, the top player will also get damaged, but I think that's nice actually lol

dhvanipa: yee thats nice, just checking my understanding of the new code

packages/world/src/systems/libraries/MoveLib.sol
vdrg: done, will merge!

PR #157, opened 2025-03-26
Call systems from programs

vdrg:
Now most functions receive a callerEntityId as the first parameters so they can be called by programs.


packages/world/src/Constants.sol
```diff
@@ -1,8 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.24;
 
+uint32 constant MAX_ENTITY_INFLUENCE_HALF_WIDTH = 10;
+
 uint32 constant MAX_PLAYER_INFLUENCE_HALF_WIDTH = 10;
```
dhvanipa, line 6: Can we remove the player one then?

packages/world/src/EntityId.sol
```diff
@@ -3,21 +3,96 @@ pragma solidity >=0.8.24;
...
+  }
+
+  function requireConnected(EntityId self, Vec3 otherCoord) internal view returns (Vec3, Vec3) {
+    Vec3 selfCoord = self.getPosition();
+    require(selfCoord.inSurroundingCube(otherCoord, MAX_ENTITY_INFLUENCE_HALF_WIDTH), "Entity is too far");
+    return (selfCoord, otherCoord);
+  }
+
+  function getPosition(EntityId self) internal view returns (Vec3) {
+    return ObjectType._get(self).isMovable() ? MovablePosition._get(self) : Position._get(self);
```
dhvanipa, line 95: In one tx ._get might be called in multiple places, eg in activate its called to check action allowed and then for position. Is there a way to somehow cache this and not do the table read (might not be possible in EVM/Solidity just a though)

packages/world/src/Constants.sol
vdrg: yeah I just left it there while replacing, but also wasn't sure if maybe we should have different influence range. Like maybe it is weird that the chest can reach up to 10 blocks?

packages/world/src/EntityId.sol
```diff
@@ -3,21 +3,96 @@ pragma solidity >=0.8.24;
...
 
 library EntityIdLib {
-  function isBaseEntity(EntityId self) internal view returns (bool) {
-    return EntityId.unwrap(BaseEntity._get(self)) == bytes32(0);
+  using ObjectTypeLib for ObjectTypeId;
+
+  function activate(EntityId self) public returns (EnergyData memory) {
+    checkWorldStatus();
+
+    ObjectTypeId objectTypeId = ObjectType.get(self);
```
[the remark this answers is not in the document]
vdrg: ye currently modifying that function a bit, ty

dhvanipa:
Overall, I like the new `EntityIdLib` utils and how they are generic beyond player! The new pattern of passing in callerEntityId makes sense. It's a bit hard for me to ensure all the logic in the systems is correct, but most are just re-naming the playerEntityId to callerEntityId so checks out

dhvanipa:
Ik the PR is in draft, but just FYI there's a couple places where notifys were commented out should bring them back before merging. Also when the PR is ready, I'll update the client in a corresponding branch

packages/world/src/EntityId.sol
dhvanipa: yea the code is easier to understand when you're not passing around stuff

dhvanipa: > could be stored in transient storage
wdym?

packages/world/src/Constants.sol
dhvanipa: oh true, it could be a util on the object type then, similar to actions allowed! eg even building drones might not have the same range

packages/world/src/EntityId.sol
vdrg: like when reading from tables you could cache them in transient storage which is cheaper than storage, but not sure if the overhead would be worth it in most cases

vdrg:
> Ik the PR is in draft, but just FYI there's a couple places where notifys were commented out should bring them back before merging. Also when the PR is ready, I'll update the client in a corresponding branch

done!

vdrg:
want to add more tests but should be ready for review

vdrg:
@dhvanipa there are some refactors I had to do to a few systems, so it would be good to review those as well

packages/world/src/EntityId.sol
```diff
@@ -3,21 +3,71 @@ pragma solidity >=0.8.24;
...
+import { MAX_ENTITY_INFLUENCE_HALF_WIDTH } from "./Constants.sol";
+
+using ObjectTypeLib for ObjectTypeId;
+
 type EntityId is bytes32;
 
 library EntityIdLib {
-  function isBaseEntity(EntityId self) internal view returns (bool) {
-    return EntityId.unwrap(BaseEntity._get(self)) == bytes32(0);
+  // We need to use this internal library function in order to obtain the msg.sig and msg.sender
```
dhvanipa, line 31: why do you need this if library functions are delegate called?

vdrg: because msg.data is not preserved in delegate calls, so you can't get the selector nor the msgSender due to how mud encodes the msg.sender in the calldata

packages/world/src/systems/TransferSystem.sol
```diff
@@ -4,142 +4,146 @@ pragma solidity >=0.8.24;
...
-        transferObjectTypeId: transferObjectTypeId,
-        transferAmount: numToTransfer
-      })
-    );
+    if (callerEntityId != fromEntityId) {
+      callerEntityId.requireConnected(fromEntityId);
+    }
+    fromEntityId.requireConnected(toEntityId);
+
+    transferEnergyToPool(callerEntityId, SMART_CHEST_ENERGY_COST);
```
dhvanipa, line 42: This energy cost would still run even if the chest doesnt have a force field right? or if the force field is out of charge, but you should still be able to take/withdraw

dhvanipa:
> @dhvanipa there are some refactors I had to do to a few systems, so it would be good to review those as well, specifically build and transfer

ayoo transfer is so clean now. Build looks like it just moved stuff around, there's no logic change from what I read

packages/world/src/Constants.sol
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.24;
 
-uint32 constant MAX_PLAYER_INFLUENCE_HALF_WIDTH = 10;
-uint32 constant MAX_PLAYER_RESPAWN_HALF_WIDTH = 10;
+uint32 constant MAX_ENTITY_INFLUENCE_HALF_WIDTH = 10;
+
```
dhvanipa, line 5: nit: can we remove this extra line xD

dhvanipa (commented):
We should rename `PlayerActionNotif` table also to just `ActionNotif`

packages/world/src/systems/TransferSystem.sol
vdrg: yes, should there be a cost for transfers? should it be drained from the forcefield instead?

packages/world/src/EntityId.sol
dhvanipa: ahh neat, yea sounds like an optional MUD thing if ever implemented

dhvanipa (approved):

packages/world/src/systems/TransferSystem.sol
dhvanipa: I was thinking initially no, but then maybe yes to be consistent with chest callers

1) player transfer items --> take energy from player
2) chest transfer items --> take energy from force field

I think 1) could technically drain the force field by just going in and out if it took from the force field, so better if it takes from the player. for 2) the force field gets to control the chests inside

dhvanipa: so then the current code is actually good as it is

PR #161, opened 2025-03-31
Refactor Program Hooks

vdrg:
Key changes (some are a bit different from what we discussed):
- `IHooks.sol` interface that defines all hooks
- New `ProgramId` type that encodes a `ResourceId`, and functions are defined for it to make calling hooks easier
- Programs are detached on mine
- Special `validateProgram` hook that is only called on **active** forcefields when attaching a program
  -  It is a called via staticcall
- Most calls to hooks revert if the hook reverts:
  - This is different from what we discussed about making them return bools
  - The reason for this is that we can make programs include an empty fallback function which will allow them to accept all hooks, effectively making hooks optional and future proof if this pattern is followed.
  - The fallback function could just return early if the world is the caller, so from the outside (when calling program functions directly) nothing changes
- The only calls to hooks that don't revert (they use a fixed amount of gas) are:
  - The call to `onDetach` if the force field the entity is in is NOT active (both in `attachProgram` and `mine`)
    - The reason this is fine vs having a different forcefield hook (like validateProgram) is that the program can easily check its forcefield's energy and decide what to do depending on that (eg if the forcefield ran out of energy, the program knows it should not revert and just do whatever cleanup it needs)
  - The call to `onHit`
- Also sneaked in the change @frolic suggested about only checking if mining is allowed for the entity being mined (if an object is partially outside a forcefield, you can mine it)
- Removed displayContent in favor of display URI
  - the `getDisplayURI` function of the DisplaySystem now also receives `extraData`: my thinking is that maybe standards could emerge where people can set a default extraData for displays in order to customize the theme or the content itself


packages/world/src/IHooks.sol
```diff
@@ -0,0 +1,80 @@
...
+import { ProgramId } from "./ProgramId.sol";
+import { Vec3 } from "./Vec3.sol";
+import { ObjectTypeId } from "./ObjectTypeId.sol";
+import { ObjectAmount } from "./ObjectTypeLib.sol";
+
+/**
+ * caller is the entity that called the system and triggered the hook
+ * target is the entity for which the hook is being called
+ */
+interface IHooks {
```
frolic, line 14: small thing but should we call this something like `IProgramHooks`? just to shy away from the built-in `Hooks.sol` lib/table in Store

I know there's no overlap right now but just wary of a potential overlap later

packages/world/src/ProgramId.sol
```diff
@@ -0,0 +1,114 @@
...
+import { EntityId } from "./EntityId.sol";
+import { Vec3 } from "./Vec3.sol";
+import { IHooks } from "./IHooks.sol";
+import { SAFE_PROGRAM_GAS } from "./Constants.sol";
+
+type ProgramId is bytes32;
+
+interface IProgramValidator {
+  function validateProgram(ProgramId newProgram) external view returns (bool);
+}
```
frolic, line 20: this seems unused?

packages/world/src/ObjectTypeLib.sol
```diff
@@ -306,7 +306,7 @@ library ObjectTypeLib {
         sig == ITransferSystem.transfer.selector ||
         sig == ITransferSystem.transferTool.selector ||
         sig == ITransferSystem.transferTools.selector ||
-        sig == IMachineSystem.powerMachine.selector;
+        sig == IMachineSystem.fuelMachine.selector;
```
frolic, line 309: does this mean "fuel source"?

packages/world/src/ProgramId.sol
```diff
@@ -0,0 +1,114 @@
...
+
+  function callOrRevert(ProgramId self, bytes memory callData) internal returns (bytes memory) {
+    (bool success, bytes memory returnData) = self.call(callData);
+    if (!success) {
+      revertWithBytes(returnData);
+    }
+    return returnData;
+  }
+
+  function safeCall(ProgramId self, bytes memory data) internal returns (bool, bytes memory) {
```
frolic, line 68: I feel like there's a movement away from the term "safe" for various call types - is there a better descriptor for what this is doing? `gasBoundedCall` or something

packages/world/src/systems/DisplaySystem.sol
```diff
@@ -42,19 +45,22 @@ contract DisplaySystem is System {
...
+      bytes memory returnData = baseEntityId.getProgram().staticcallOrRevert(_getDisplayURI);
+      return abi.decode(returnData, (string));
     }
 
-    return DisplayContentData({ contentType: DisplayContentType.None, content: bytes("") });
+    return "";
   }
 
-  function setDisplayContent(EntityId entityId, DisplayContentData memory content) public {
+  function setDisplayURI(EntityId /* caller */, EntityId entityId, string memory uri) public {
```
frolic, line 56: is this needed anymore now that `getDisplayURI` is a getter?

vdrg: Yes, want to sync with dvhani about that, not sure if we still want to support setting display data without a program

packages/world/src/ObjectTypeLib.sol
vdrg: from my understanding all entities that have their own fuel tank are called "machines" (forcefields, drones, etc)

packages/world/src/IHooks.sol
vdrg: ye I agree, I just think IHooks is more comfortable when using it because it is a little bit shorter. Maybe we could call them events? or come up with some other way for calling hooks.

Originally I had directly implemented the hooks in the ProgramId lib so you could do `program.onMine(...)`, but I didn't like that the same hook was defined in the interface (to be able to use encodeCall) and in the library. Also most hooks are only used in a single place, so adding it to the library didn't make much sense. And that meant that the way they were called (safeCall, staticcall, etc) was done in the lib as well so it felt like too much abstraction and less control within the systems that call them.

packages/world/src/ProgramId.sol
vdrg: yeah I kind of agree, I'll try to come up with something better

packages/world/src/ProgramId.sol
vdrg: hadn't saved the file 🤦

packages/world/src/ProgramId.sol
vdrg: Ok changed it to:

```solidity
    program.call({ gas: SAFE_PROGRAM_GAS, hook: onHit });
```

That way it is more explicit. Lmk what you think

packages/world/src/IHooks.sol
vdrg: Yeah validateProgram and getDisplayURI don't feel like hooks, so we could also move them to a different interface. I kinda like IProgram, but could be confusing as it sounds like it is something your program has to inherit from?

frolic: what about `abstract contract Program` and implement the hooks but revert? like https://github.com/latticexyz/mud/blob/main/packages/world/src/Module.sol

frolic: what is this interface solving for? so you can do `IHooks(program).someHook()`?

packages/world/src/ObjectTypeLib.sol
vdrg: not all smart entities can be refueled though, so I believe we still need a different term for those

packages/world/src/IHooks.sol
vdrg: This also solves the issue with naming, ie hooks can be called ISleepHook.onSleep, and non-hooks can be IProgramValidator.validateProgram, IDisplay.getDisplayURI, etc

packages/world/src/IHooks.sol
```diff
@@ -0,0 +1,80 @@
...
+  function validateProgram(
+    EntityId caller,
+    EntityId target,
+    EntityId programmed,
+    ProgramId program,
+    bytes memory extraData
+  ) external;
+
+  // TODO: should we move view functions somewhere else?
+  function getDisplayURI(EntityId caller, EntityId target, bytes memory extraData) external returns (string memory);
```
dhvanipa, line 24: This should be marked as view right?

vdrg: About to push some changes!

packages/world/src/ObjectTypeLib.sol
dhvanipa: yee smart entities is more general, since chests and doors are also smart entities like force fields that have programs

machines = smart entities that can have energy
fuel = consumed by machines to increase their energy
food = consumed by players to increase their energy

packages/world/src/systems/BedSystem.sol
```diff
@@ -87,10 +90,12 @@ contract BedSystem is System {
     bedEntityId = bedEntityId.baseEntityId();
     require(!BedPlayer._getPlayerEntityId(bedEntityId).exists(), "Bed full");
 
+    // TODO: do we need a forcefield for sleeping?
```
dhvanipa, line 93: I think it makes sense to only allow sleeping in charged force fields. If we dont require a charged force field for sleeping, where would the player energy drain from?

packages/world/src/IHooks.sol
```diff
@@ -0,0 +1,80 @@
...
+    bytes memory extraData
+  ) external;
+
+  // Machines
+  function onHit(EntityId caller, EntityId target, uint128 damage, bytes memory extraData) external;
+
+  function onFuel(EntityId caller, EntityId target, uint16 fuelAmount, bytes memory extraData) external;
+
+  // Forcefield
+  function onAddFragment(EntityId caller, EntityId target, EntityId added, bytes memory extraData) external;
```
dhvanipa, line 47: I like these new fragment names over expand and contract!

packages/world/src/ProgramId.sol
```diff
@@ -0,0 +1,94 @@
...
+  function staticcallOrRevert(ProgramId self, bytes memory callData) internal view returns (bytes memory) {
+    (bool success, bytes memory returnData) = self.staticcall(callData);
+    if (!success) {
+      revertWithBytes(returnData);
+    }
+    return returnData;
+  }
+
+  function _hookContext(bytes memory hook) private pure returns (bytes memory) {
+    return WorldContextProviderLib.appendContext({ callData: hook, msgSender: address(0), msgValue: 0 });
```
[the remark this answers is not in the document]
vdrg: The thing that worries me about this is that mud appends msgSender and msgValue to the calldata, so if you are calling a hook within a public library the calldata is not preserved. I'm down for supporting eth transfers it's just that we have to be really careful otherwise people might be able to craft malicious calls (by passing extradata to system calls) that then would be extracted as msgSender and msgValue

packages/world/src/systems/ProgramSystem.sol
```diff
@@ -1,157 +1,120 @@
...
+    bytes memory validateProgram = abi.encodeCall(
+      IProgramValidator.validateProgram,
+      (caller, validator, target, program, extraData)
+    );
 
-    // ForceField fragments don't have a position on the grid, so we need to handle them differently
-    if (objectTypeId == ObjectTypes.ForceFieldFragment) {
-      // TODO: figure out proximity checks for fragments
-    }
+    // The validateProgram view function should revert if the program is not allowed
```
dhvanipa, line 51: Why is the validateProgram a view function?

packages/world/src/systems/ProgramSystem.sol
```diff
@@ -1,157 +1,120 @@
...
-  function attachProgram(
-    EntityId callerEntityId,
-    EntityId targetEntityId,
-    ResourceId programSystemId,
-    bytes calldata extraData
-  ) public payable {
-    callerEntityId.activate();
-    (, Vec3 targetCoord) = callerEntityId.requireConnected(targetEntityId);
+    (, bool publicAccess) = Systems._get(program.toResourceId());
+    require(!publicAccess, "Program system must be private");
```
dhvanipa, line 42: I assume this wont block programs having functions that can be called outside the world because it can be a separate program system? eg setPrices and onTransfer are in different systems?

packages/world/src/systems/ProgramSystem.sol
vdrg: so that we prevent reentrant calls, and onAttach will be called afterwards anyways so you can do state changes there. Unless we want forcefield to modify state when validating a program, but it felt more natural like this

packages/world/src/systems/ProgramSystem.sol
vdrg: private systems can still be called by systems in the same namespace. From mud docs:

> A private System can only be called by accounts that have access. This access can be the result of:
> - Access permission to the namespace in which the System is registered.
> - Access permission specifically to the System.

packages/world/src/systems/BedSystem.sol
vdrg: sg, let's add on a different pr, will create an issue

vdrg: another option would be to just warn the user instead I guess, that way we keep things simple

packages/world/src/ProgramId.sol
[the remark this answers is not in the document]
dhvanipa: ahh I see, we dont need to support ETH transfers if security is a concern lol

dhvanipa (approved):
smart items framework lfg

PROGRAMS.md
```diff
@@ -0,0 +1,284 @@
...
+- **Taxation Systems**: Automated collection and distribution of resources
+
+## 6. Implementation Guide
+
+### 6.1 Creating a Program (WIP)
+
+Example of a simple door program:
+
+```solidity
+contract SimpleDoorProgram is Program {
```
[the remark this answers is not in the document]
dhvanipa: eg right now I have to figure out which hooks to add for a bed/force field. vs I could just do `is ForceFieldProgram`, and then I'd know the ones that I have to implement that will get called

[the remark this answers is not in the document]
vdrg: It only exists in our imagination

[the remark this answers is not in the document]
vdrg: I think we should eventually have a "dust standard library" for everything related to contracts that needs to be imported by users, we can include some utilities and a base Program contract with some helpers

packages/world/src/systems/DisplaySystem.sol
vdrg: we discussed at some point that adding displays would increase the forcefield's energy drain rate. Maybe the client could filter specific uri schemas if no forcefield? like only allow static text if no forcefield. But then it wouldn't really be part of the physics

dhvanipa: yea one thing about displays is that what they show is not really part of the physics, cuz clients can render whatever they want lol

PROGRAMS.md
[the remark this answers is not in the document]
vdrg: > eg right now I have to figure out which hooks to add for a bed/force field. vs I could just do is ForceFieldProgram, and then I'd know the ones that I have to implement that will get called

Sorry I missed this one, the thing is that all hooks should be optional right? like maybe you don't want to implement wakeup, just sleep. But yeah I guess with this design you could also provide base programs that group related hooks together, though I would still include the "default" fallback in them so that new hooks can be added in the future.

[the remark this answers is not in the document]
dhvanipa: yee they're already optional cuz the physics doesnt require them. I mean in the "standard library" which is outside the physics it could have a base force field program contract, which is a guide to the user of which methods to implement. Otherwise as a user I have to figure that out myself

packages/world/src/ProgramId.sol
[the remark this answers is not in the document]
vdrg: let's leave it for later, we can think about how to do it safely

PROGRAMS.md
```diff
@@ -0,0 +1,223 @@
+# WIP Programs: A Comprehensive Analysis
+
+## Abstract
+
+This document provides a detailed examination of the Programs architecture within the DUST smart contract ecosystem. Programs represent a powerful abstraction layer that enables dynamic behavior and programmable logic for entities within the virtual world. By leveraging the MUD framework's capabilities, Programs offer a flexible mechanism for extending entity functionality through custom logic that can respond to various in-world events. This paper explores the theoretical foundations, implementation details, and practical applications of Programs, serving both as an academic analysis and a practical guide for developers.
```
alvrs, line 5: I know this is not meant to be final but here are a few suggestions to make it sound a bit more down to earth haha
```suggestion
This document provides an overview of the Programs architecture within the DUST smart contract ecosystem. Programs represent a powerful abstraction layer that enables dynamic behavior and programmable logic for entities within the virtual world. By leveraging the MUD framework's capabilities, Programs offer a flexible mechanism for extending entity functionality through custom logic that can respond to various in-world events. This document explains the concepts, implementation details and practical applications of Programs, serving as a practical guide for developers.
```

PROGRAMS.md
```diff
@@ -0,0 +1,223 @@
...
+
+Programs function as an extension mechanism that separates an entity's data from its behavior, following principles similar to the Entity Component System (ECS) pattern. This separation enables dynamic composition of functionality and promotes code reuse across different entity types.
+
+## 2. Framework
+
+### 2.1 Core Concepts
+
+Programs in this architecture are defined by several key abstractions:
+
+- **ProgramId**: A unique identifier that references a specific program implementation (the underlying value is a MUD's `ResourceId`)
```
alvrs, line 19:
```suggestion
- **ProgramId**: A unique identifier that references a specific program implementation (the underlying value is a MUD `ResourceId`)
```

PROGRAMS.md
```diff
@@ -0,0 +1,223 @@
...
+}
+```
+
+This allows machines to execute custom logic when they receive fuel, potentially triggering complex behaviors or state changes.
+
+## 5. Practical Applications
+
+### 5.1 Smart Objects
+
+Programs enable the creation of smart objects with custom behaviors:
```
alvrs, line 131:
```suggestion
Programs enable the creation of smart entities with custom behaviors:
```

PROGRAMS.md
```diff
@@ -0,0 +1,223 @@
...
+  // Notification logic
+  // ...
+}
+```
+
+This allows machines to execute custom logic when they receive fuel, potentially triggering complex behaviors or state changes.
+
+## 5. Practical Applications
+
+### 5.1 Smart Objects
```
alvrs, line 129:
```suggestion
### 5.1 Smart Entities
```

PROGRAMS.md
```diff
@@ -0,0 +1,223 @@
...
+
+### 6.1 Creating a Program (WIP)
+
+Example of a simple door program:
+
+```solidity
+contract SimpleDoorProgram is Program {
+  constructor(IWorld world) Program(world) {}
+
+  function onAttach(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
```
alvrs, line 166: nit: was there a decision around renaming `extraData` to something more descriptive? (ie `programContext` or sth?)
```suggestion
  function onAttach(EntityId caller, EntityId target, bytes memory programContext) external onlyWorld {
```

packages/world/src/systems/ProgramSystem.sol
```diff
@@ -1,157 +1,120 @@
...
-  function _requireInterface(address programAddress, bytes4 interfaceId) internal view {
-    require(
-      ERC165Checker.supportsInterface(programAddress, interfaceId),
-      "Program does not implement the required interface"
-    );
-  }
+    // ForceField fragments don't have a position on the grid, so we need to handle them differently
+    // if (objectTypeId == ObjectTypes.ForceFieldFragment) {
+    //   // TODO: figure out proximity checks for fragments
+    // }
```
alvrs, line 37: can this be removed?

packages/world/src/systems/ProgramSystem.sol
```diff
@@ -1,157 +1,120 @@
...
+    caller.activate();
+    (, Vec3 targetCoord) = caller.requireConnected(target);
+    target = target.baseEntityId();
 
     // ForceField fragments don't have a position on the grid, so we need to handle them differently
-    if (ObjectType._get(baseEntityId) == ObjectTypes.ForceFieldFragment) {
-      // TODO: figure out proximity checks for fragments
+    // if (ObjectType._get(baseEntityId) == ObjectTypes.ForceFieldFragment) {
+    //   // TODO: figure out proximity checks for fragments
+    // }
```
alvrs, line 69: can this be removed?

alvrs (approved):
really like how this turned out!

PROGRAMS.md
vdrg: oh yeah this was mostly AI generated, just so we can at least have something, will include your changes

PROGRAMS.md
vdrg: no decision but yeah I think we could do it, will leave for another pr so we can merge this soon

PR #165, opened 2025-04-02
Add/Remove single fragments

vdrg:
Simplified forcefield system to only support adding/removing single fragments on each call


vdrg:
closing this, will create a new branch from main and apply the changes

PR #168, opened 2025-04-02
forge fmt

frolic:
way faster than prettier and very apparent in pre-commit hook and compared with biome for JS/TS formatting

before:
image

after:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
   "typescript.updateImportsOnFileMove.enabled": "always",
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "solidity.formatter": "prettier",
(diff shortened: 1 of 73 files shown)


vdrg (approved):
nice! the only thing I'm not sure about is sorting the imports. Currently imports are kind of a mess, I agree with that and we should clean them and follow some convention. However, I like when imports go from external libraries -> local files. OTOH I prefer having just alphabetical sorting to not having any convention.

frolic:
> nice! the only thing I'm not sure about is sorting the imports. Currently imports are kind of a mess, I agree with that and we should clean them and follow some convention. However, I like when imports go from external libraries -> local files. OTOH I prefer having just alphabetical sorting over not having a convention at all.

agree on external -> local, might be a good feature request for foundry

dhvanipa:
> agree on external -> local, might be a good feature request for foundry

+1 I like external --> local

PR #169, opened 2025-04-02
publish package preview

frolic:
see https://github.com/stackblitz-labs/pkg.pr.new

### DIFF
--- a/.github/workflows/publish-preview.yml
+++ b/.github/workflows/publish-preview.yml
@@ -6,8 +6,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
       - run: corepack enable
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-preview.yml
```diff
@@ -0,0 +1,24 @@
...
+          node-version: 20
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm build
+
+      - run: pnpm dlx pkg-pr-new publish --comment=off --pnpm './packages/dustkit'
```
frolic, line 24: should we also start publishing a preview of the world package?

dhvanipa: yea we should. Is there a cost to using this service? or its free?

frolic: freeee babyyyy

PR #175, opened 2025-04-02
Balancing first pass

dhvanipa:
- Set mass and energy values from excel
- Apply scale factor
- Remove smart item versions
- Renamed silver ore to iron ore and neptunium ore to emerald ore


packages/world/src/Constants.sol
```diff
@@ -23,26 +23,27 @@ uint256 constant RESPAWN_ORE_BLOCK_RANGE = 10;
...
-uint128 constant HIT_ENERGY_COST = 50;
-uint128 constant TILL_ENERGY_COST = 50;
-uint128 constant CRAFT_ENERGY_COST = 50;
-uint128 constant MOVE_ENERGY_COST = 5;
+uint128 constant BUILD_ENERGY_COST = 81 * SCALE_FACTOR;
+uint128 constant MINE_ENERGY_COST = 81 * SCALE_FACTOR;
+uint128 constant HIT_ENERGY_COST = 81 * SCALE_FACTOR;
+uint128 constant TILL_ENERGY_COST = 81 * SCALE_FACTOR;
+uint128 constant CRAFT_ENERGY_COST = 81 * SCALE_FACTOR;
+uint128 constant MOVE_ENERGY_COST = 0.2554375 * 10 ** 14;
```
vdrg, line 39: do we want to just hardcode the values so it is consistent? it is weird that some use scale factor and some don't

packages/world/src/systems/MineSystem.sol
```diff
@@ -276,18 +276,13 @@ library RandomOreLib {
...
-      ObjectTypes.GoldOre,
-      ObjectTypes.DiamondOre,
-      ObjectTypes.NeptuniumOre
-    ];
-
-    uint256[5] memory max = [MAX_COAL, MAX_SILVER, MAX_GOLD, MAX_DIAMOND, MAX_NEPTUNIUM];
-
-    // For y > -50: More common ores (coal, silver)
-    // For y <= -50: More rare ores (gold, diamond, neptunium)
+    ObjectTypeId[5] memory ores =
```
vdrg, line 279: did you run forge fmt 👀  ?

packages/world/test/Bed.t.sol
```diff
@@ -369,7 +369,7 @@ contract BedTest is DustTest {
 
     // After 1000 seconds, the forcefield should be depleted
     // We wait more time so the player's energy is FULLY depleted in this period
-    uint128 playerDrainTime = initialPlayerEnergy / PLAYER_ENERGY_DRAIN_RATE;
+    uint128 playerDrainTime = initialPlayerEnergy / PLAYER_ENERGY_DRAIN_RATE + 1;
```
vdrg, line 372: this makes sense I think but maybe a comment explaining the + 1?

vdrg (approved):
lgtm! small comments

packages/world/src/systems/MineSystem.sol
dhvanipa: its auto running for me in vscode ye

vdrg (approved):

PR #176, opened 2025-04-03
feat: farming

alvrs:


packages/world/src/timeToGrow.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+/* Autogenerated file. Do not edit manually. */
+
+import { ObjectTypeId } from "./ObjectTypeId.sol";
+import { ObjectTypes } from "./ObjectTypes.sol";
+
+function timeToGrow(ObjectTypeId objectTypeId) pure returns (uint128) {
+  if (objectTypeId == ObjectTypes.WheatSeed) {
```
dhvanipa, line 10: Why doenst this have the tree seeds?

dhvanipa: also can we make this 15 minutes? and then use scripts to speed up time when doing dev locally

dhvanipa (approved):

PR #178, opened 2025-04-03
Random seed drops 🌱 

### LINKED ISSUE #140: Implement randomness for seeds/crops
- Similar to ores, have a cap for each seed, probabilities depend on these
- You can also get nothing with certain probability
- When harvesting a crop, there's a probability to also get a seed
- if you don't get a seed, it gets added back to the total supply

vdrg:
Fixes #140


README.md
```diff
@@ -1,21 +0,0 @@
...
-
-```
-pnpm run dev
-```
-
-### Installation
-
-Make sure sure you have:
-
-- [Node.js v18](https://nodejs.org/en/download/package-manager)
```
dhvanipa, line 15: We should still keep the installation and run instructions somewhere. Prob in the same/new README under `packages/world`

packages/world/README.md
```diff
@@ -0,0 +1,174 @@
...
+- For example, Wheat has a chance to drop 0-3 seeds
+- Base probabilities (for n=3, p=0.57):
+  - 0 seeds: ~8%
+  - 1 seed: ~31%
+  - 2 seeds: ~41%
+  - 3 seeds: ~20%
+- These probabilities are adjusted based on remaining capacity
+- For multiple seeds, a compound probability is calculated
+
+## Implementation Details
```
dhvanipa, line 94: great doc!!

packages/world/src/Constants.sol
```diff
@@ -38,6 +38,11 @@ uint128 constant CRAFT_ENERGY_COST = 8100000000000000;
 uint128 constant MOVE_ENERGY_COST = 25543750000000;
 uint128 constant PLAYER_FALL_ENERGY_COST = MAX_PLAYER_ENERGY / 25; // This makes it so, with full energy, you die from a 25 block fall
 
+// Resource caps
+uint256 constant MAX_WHEAT_SEED = 5_000_000;
+uint256 constant MAX_OAK_SEED = 1_000_000;
```
dhvanipa, line 43: Are these from the new balancing sheet or just set arbitrarily for now?

README.md
vdrg: yes! removed by mistake

packages/world/src/Constants.sol
vdrg: arbitrary for now

packages/world/src/systems/MineSystem.sol
```diff
@@ -113,15 +113,15 @@ contract MineSystem is System {
...
+      mineObjectTypeId = RandomResourceLib._collapseRandomOre(entityId, coord);
     }
 
+    uint128 finalMass = MassReductionLib._processMassReduction(caller, entityId);
+
     // First coord will be the base coord, the rest is relative schema coords
     Vec3[] memory coords = mineObjectTypeId.getRelativeCoords(baseCoord, Orientation._get(entityId));
 
-    uint128 finalMass;
     {
```
dhvanipa, line 124: can these curly braces be removed now? it doesnt seem like the scope has no variable instantiations anymore

packages/world/src/NatureLib.sol
```diff
@@ -0,0 +1,230 @@
...
+    uint256 totalWeight = 0;
+    for (uint256 i = 0; i < weights.length; i++) {
+      totalWeight += weights[i];
+    }
+
+    require(totalWeight > 0, "No options available");
+
+    // Initialize PRNG
+    // TODO: maybe this is not necessary as we already have a random seed
+    LibPRNG.PRNG memory prng;
```
dhvanipa, line 173: we still need to call `seed` on `prng` though right? or by this todo do you mean we could not use `uniform`

packages/world/src/Constants.sol
dhvanipa: created an issue to track: https://github.com/dustproject/dust/issues/181

dhvanipa (approved):
really like how things are generalized for random resources now. Will be easy to introduce new ones, great workk 🎉

PR #179, opened 2025-04-03
Fix placeable positions

dhvanipa:
The client has a limitation on which directions it can support rendering, see PR description: https://github.com/dustproject/dust-client/pull/655

On the contracts therefore we limit the directions for beds. Otherwise, they'd exist in the world but our client wouldn't be able to or would incorrectly render them.


vdrg (approved):
makes sense as a temporary patch

dhvanipa:
I have a different approach in mind that doesnt require a hard code for beds, but actually doing it in a separate PR

Issue #149, opened 2025-03-20
Make tool reduction a percentage of max tool mass


vdrg:

vdrg:
Fixed by #182

PR #182, opened 2025-04-06
Inventory refactor

vdrg:
I want to keep improving the code but don't want to block wiresaw integration, rn there are a few things that can be made more efficient:
- [x] Directly replacing bucket types by passing a slot when watering farmland
- [ ] Simplify crafting by passing slots of the inputs and amounts (I think this is also useful to users so they can decide which items to use for recipes)
- [ ] The new inventory utils can definitely be improved. Right now there is some (maybe unnecessary) complexity related to keeping track of the slots for each object type:
  -  One option is to just simplify everything and remove tracking type slots, but operations like `transferAll`, pickup on mining, etc could become more expensive as we would need to iterate over the whole inventory
  - The other option I was considering but is more complex is to split the type slots into partially and fully filled slots, which could make some operations cheaper but it is unclear if the overhead of tracking these slots could counter the gains in efficiency


packages/world/src/systems/DropSystem.sol
```diff
@@ -1,68 +0,0 @@
...
-import { ObjectTypes } from "../ObjectTypes.sol";
-import { getUniqueEntity } from "../Utils.sol";
-import { Vec3 } from "../Vec3.sol";
-import { getOrCreateEntityAt } from "../utils/EntityUtils.sol";
-import { transferInventoryEntity, transferInventoryNonEntity } from "../utils/InventoryUtils.sol";
-import { DropNotification, notify } from "../utils/NotifUtils.sol";
-import { TerrainLib } from "./libraries/TerrainLib.sol";
-
-// TODO: combine the tool and non-tool drop functions
-contract DropSystem is System {
```
dhvanipa, line 23: Why was DropSystem removed?

vdrg: I moved both drop and pickup to inventory system, we were testing both of them together anyways, but can revert

dhvanipa: wait nvm spoke too soon I see `InventorySystem`

packages/world/src/systems/MineSystem.sol
```diff
@@ -165,35 +176,44 @@ contract MineSystem is System {
...
+  function mineUntilDestroyed(EntityId caller, Vec3 coord, uint16 toolSlot, bytes calldata extraData) public payable {
+    uint128 massLeft = 0;
+    do {
+      // TODO: factor out the mass reduction logic so it's cheaper to call
+      EntityId entityId = _mine(caller, coord, toolSlot, extraData);
+      massLeft = Mass._getMass(entityId);
+    } while (massLeft > 0);
   }
 
   function mineUntilDestroyed(EntityId caller, Vec3 coord, bytes calldata extraData) public payable {
```
dhvanipa, line 212: We should do the same for `mine` and `mineUntilDestroyed` for tools and not tools. Either use overloaded functions or add a suffix (vs doing both) for consistency

packages/world/src/utils/ReadUtils.sol
```diff
@@ -53,34 +49,34 @@ struct EntityData {
...
-          inventoryEntities[k] = InventoryEntity({ entityId: tool, mass: Mass._getMass(tool) });
-          k++;
-        }
-      }
-    }
-    inventoryObjects[i] =
-      InventoryObject({ objectTypeId: objectTypeId, numObjects: count, inventoryEntities: inventoryEntities });
-  }
-  return inventoryObjects;
+  // uint16[] memory objectTypeIds = InventoryObjects._get(entityId);
```
dhvanipa, line 52: oh hmm, with this removed the client wont be able to sync inventory cuz its using the getter for now, so this will break mine and builds

Is there a way to have this getter still work @vdrg (even if its gas expensive)?

Otherwise, @alvrs will the change to indexer to use slots be easy or take a while? wondering if we should merge this in and have broken inventory or wait

vdrg: ah yeah I can just iterate over all the slots and get the types

packages/world/src/systems/admin/AdminSystem.sol
```diff
@@ -37,27 +34,17 @@ contract AdminSystem is System {
...
   }
 
   function adminAddToolToInventory(EntityId owner, ObjectTypeId toolObjectTypeId) public onlyAdmin returns (EntityId) {
-    return addToolToInventory(owner, toolObjectTypeId);
+    EntityId tool = createEntity(toolObjectTypeId);
+    InventoryUtils.addEntity(owner, tool);
+    return tool;
   }
 
-  function adminRemoveFromInventory(EntityId owner, ObjectTypeId objectTypeId, uint16 numObjectsToRemove)
```
dhvanipa, line 47: Why were the remove utils removed?

packages/world/src/utils/ReadUtils.sol
dhvanipa: okay sweet yea then at least it will still work even if its expensive

packages/world/src/systems/admin/AdminSystem.sol
vdrg: ah yeah meant to add those back later, can do it now with the new utils

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -1,244 +1,404 @@
...
-  require(stackable > 0, "This object type cannot be removed from the inventory");
+    toolType = ObjectType._get(tool);
+    require(toolType.isTool(), "Inventory item is not a tool");
+    uint128 toolMassLeft = Mass._getMass(tool);
+    require(toolMassLeft > 0, "Tool is already broken");
 
-  uint16 numInitialFullStacks = numInitialObjects / stackable;
-  bool hasInitialPartialStack = numInitialObjects % stackable != 0;
+    uint128 maxMass = ObjectTypeMetadata._getMass(toolType);
+    uint128 maxUsePerCall = maxMass / 10; // Limit to 10% of max mass per use
```
dhvanipa, line 46: niceee, so this fixes https://github.com/dustproject/dust/issues/149

vdrg: yes! wanted to confirm if that is the exact thing we want

dhvanipa (approved):
okay I understand `InventoryUtils` now which is the main code change. The complexity of the two indexes made it a bit hard to understand but dont think we should block merging for wiresaw changes

I dont think we're going to not use slots because of wiresaw, but I'm curious what the gas difference is with slots vs without (ie whats on main right now)

packages/world/src/utils/InventoryUtils.sol
dhvanipa: yup thats how I was thinking of it

packages/world/src/systems/MineSystem.sol
vdrg: done using overloaded functions, let me know if I should just use different names

dhvanipa: can keep this for now and do different names if its causing issues

PR #184, opened 2025-04-07
Add default programs

dhvanipa:
- See README under `packages/programs`
- Added a deploy script that allows us to run `pnpm run deploy:local` from the new programs package and have it deploy to the world from the worlds package (by default it would deploy to a new world or you'd have to copy the world address manually)


packages/programs/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+bindings/
+artifacts/
+broadcast/
+
+# Ignore MUD deploy artifacts
+deploys/**/*.json
+dist/
+
+# We ignore this file because we use the one in the world package as the source of truth
+worlds.json
```
dhvanipa, line 13: I added this so we dont have two worlds.json which would be confusing

packages/programs/deployToWorld.bash
frolic, line 1: can we use mprocs or npn scripts

packages/programs/mud.config.ts
```diff
@@ -0,0 +1,54 @@
+import { defineWorld } from "@latticexyz/world";
+
+export default defineWorld({
+  // Note: this is required as the world is deployed with this
```
frolic, line 4: Even when passing world address?

packages/programs/script/PostDeploy.s.sol
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    // Set the default programs
+    bytes32[] memory defaultPrograms = new bytes32[](3);
+    defaultPrograms[0] = ffProgramId.unwrap();
+    defaultPrograms[1] = chestProgramId.unwrap();
+    defaultPrograms[2] = bedProgramId.unwrap();
+    DefaultPrograms.set(defaultPrograms);
+
+    vm.stopBroadcast();
```
frolic, line 65: if for now we name these like "BedProgramSystem" we don't need to manually deploy them 

Can go back to "BedProgram" once we have the MUD PR landed that opens up naming systems

packages/programs/mud.config.ts
```diff
@@ -0,0 +1,54 @@
...
+  deploy: {
+    upgradeableWorldImplementation: true,
+  },
+  namespace: "default-1",
```
vdrg, line 8: should we come up with some convention for versioning these?

packages/programs/mud.config.ts
```diff
@@ -0,0 +1,54 @@
...
+      filePath: "@latticexyz/store/src/ResourceId.sol",
+      type: "bytes32",
+    },
+    ProgramId: {
+      filePath: "@dust/world/src/ProgramId.sol",
+      type: "bytes32",
+    },
+  },
+  tables: {
+    Admin: {
```
vdrg, line 25: wdyt of calling it `Owner` seems more standard in the smart contract world

vdrg: while admin feels like something more global, like for the shop program there will be an admin that can set the global fee and fee recipient

packages/programs/mud.config.ts
```diff
@@ -0,0 +1,54 @@
...
+    AllowedPlayers: {
+      schema: {
+        entityId: "EntityId",
+        players: "address[]",
+      },
+      key: ["entityId"],
+    },
+    DefaultPrograms: {
+      schema: {
+        programs: "bytes32[]", // ProgramId[]
```
vdrg, line 41: is this mapping necessary? can't we just keep it in the client?

packages/programs/src/programs/DefaultProgram.sol
```diff
@@ -0,0 +1,62 @@
...
+contract DefaultProgram is IAttachProgramHook, IDetachProgramHook, WorldConsumer {
+  /**
+   * @notice Initializes the DefaultProgram
+   * @param _world The world contract
+   */
+  constructor(IBaseWorld _world) WorldConsumer(_world) { }
+
+  function onAttachProgram(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
+    address player = ReversePlayer.get(caller);
+    require(player != address(0), "Caller is not a player");
```
vdrg, line 27: hmm I wonder if we should just use owner entityIds directly, as maybe some other entity might want to use these programs?

packages/programs/mud.config.ts
```diff
@@ -0,0 +1,54 @@
...
+      schema: {
+        entityId: "EntityId",
+        admin: "address",
+      },
+      key: ["entityId"],
+    },
+    AllowedPlayers: {
+      schema: {
+        entityId: "EntityId",
+        players: "address[]",
```
vdrg, line 35: wdyt of having `(smart entity, caller) => bool allowed` plus a counter? then you can set allowed entities individually vs having to send the whole array

packages/programs/src/programs/ForceFieldProgram.sol
```diff
@@ -0,0 +1,98 @@
...
+  function validateProgram(
+    EntityId caller,
+    EntityId target,
+    EntityId programmed,
+    ProgramId program,
+    bytes memory extraData
+  ) external view onlyWorld {
+    bytes32[] memory defaultPrograms = DefaultPrograms.get();
+    for (uint256 i = 0; i < defaultPrograms.length; i++) {
+      if (defaultPrograms[i] == program.unwrap()) {
```
vdrg, line 55: I see now the need for default programs table, but this one is kind of weird, I don't think the default forcefield should just allow default programs, a saner default imo is to accept any program

vdrg: and maybe allow restricting to specific programs with another function

vdrg (commented):
looks good but not sure about the table schema and default programs

packages/programs/mud.config.ts
dhvanipa: one problem is how to do cleanup?

`onDetachProgram`, we wouldnt be able to clean up the old allowed values. If someone then places a new smart item in the same position, they'd get the same entity id (since its tied to position), and it would incorrectly have the previous approved callers

packages/programs/src/programs/ForceFieldProgram.sol
dhvanipa: for now removing restricting all together. default can just approve any program yea

packages/programs/mud.config.ts
vdrg: with the counter you can iterate through all of them, or you can do it as in the shop program where there is an indirection chest => shopId => shop info.

shopId is set incrementally on attachment, so it is always unique and you just use a new shop

packages/programs/mud.config.ts
dhvanipa: ye can use a convention, used `-1` as I saw it in `mock-apps`

packages/programs/mud.config.ts
dhvanipa: renamed!

packages/programs/src/programs/DefaultProgram.sol
dhvanipa: made it use entity ids

packages/programs/script/PostDeploy.s.sol
dhvanipa: Its not compiling when I rename any of them to `BedProgramSystem`

```
(node:63352) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
null
/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:3252
    }, this) : t2.accept(this);
                  ^

TypeError: Cannot read properties of null (reading 'accept')
    at ASTBuilder.visit (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:3252:19)
    at ASTBuilder.visitContractPart (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:37083:17)
    at ContractPartContext.accept (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:33779:22)
    at ASTBuilder.visit (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:3252:19)
    at /Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:37093:58
    at Array.map (<anonymous>)
    at ASTBuilder.visitContractDefinition (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:37093:41)
    at ContractDefinitionContext.accept (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:33694:22)
    at ASTBuilder.visit (/Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:3252:19)
    at /Users/dhvanipa/Documents/Tenet/AW/dust/node_modules/.pnpm/@solidity-parser+parser@0.17.0/node_modules/@solidity-parser/parser/dist/index.cjs.js:37076:59

Node.js v20.10.0
```

packages/programs/mud.config.ts
dhvanipa: ahh nvm it works fine, removed it

packages/programs/deployToWorld.bash
dhvanipa: updated

dhvanipa:
> looks good but not sure about the table schema and default programs

I revamped the table schema and programs. It gen's a new id on attach so dont actually need to do any cleanup in detach. Let me know what you think!

packages/programs/src/programs/DefaultProgram.sol
```diff
@@ -0,0 +1,53 @@
...
+import { AllowedCaller } from "../codegen/tables/AllowedCaller.sol";
+import { Owner } from "../codegen/tables/Owner.sol";
+
+import { SmartItem } from "../codegen/tables/SmartItem.sol";
+import { UniqueEntity } from "../codegen/tables/UniqueEntity.sol";
+
+contract DefaultProgram is IAttachProgramHook, IDetachProgramHook, WorldConsumer {
+  constructor(IBaseWorld _world) WorldConsumer(_world) { }
+
+  function _requireOwner(EntityId target) internal view returns (bytes32) {
```
vdrg, line 20: nit: I'm beginning to move these internal functions to the bottom of the file, to follow the solidity style guide

packages/programs/src/programs/DefaultProgram.sol
```diff
@@ -0,0 +1,53 @@
...
+    bytes32 smartItemId = SmartItem.get(target);
```
vdrg, line 21: nit: I would pass the smartItemId to this function instead of getting it here, so the return value can be removed, makes the requireOwner function more focused on just doing what it is supposed to do

packages/programs/src/programs/DefaultProgram.sol
```diff
@@ -0,0 +1,53 @@
...
+    UniqueEntity.set(uniqueEntity);
+    bytes32 smartItemId = bytes32(uniqueEntity);
+    SmartItem.set(target, smartItemId);
+
+    Owner.set(smartItemId, caller);
+    AllowedCaller.set(smartItemId, caller, true);
+  }
+
+  function onDetachProgram(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
+    require(Owner.get(SmartItem.get(target)) == caller, "Only the owner can detach this program");
```
vdrg, line 41: I would still do `deleteRecord` to smartItemId so that it gets indexed

vdrg: and also maybe remove the require if the forcefield it is in has no energy, so it can perform the cleanup on detach

vdrg (approved):
lgtm! just some small optional things

packages/programs/script/PostDeploy.s.sol
vdrg: was this solved?

packages/programs/README.md
```diff
@@ -0,0 +1,22 @@
+# Default Programs
+
+The Dust client auto selects a default program when a player builds a smart object.
+
+Default programs will make the builder the owner of that smart object. As an owner they are the only ones allowed to manage it, but they can change the owner to someone else (or to address(0) to freeze it).
```
vdrg, line 5: will we transfer the ownership of the default program namespace? or will we keep it until things are more stable?

packages/programs/script/PostDeploy.s.sol
dhvanipa: no I couldnt figure out why its failing, just renaming the contract & file to this causes the error

packages/programs/src/programs/DefaultProgram.sol
dhvanipa: Added the `deleteRecord`, didnt add the require for force field because I realized we cant use the `getForceField` so it's gonna be a lot of new code to copy. This also seems like a common util all programs will use so should add it somewhere accessible, ie that other programs can use as well

Added a todo for now

packages/programs/README.md
dhvanipa: we'll keep it until things are stable

packages/programs/script/PostDeploy.s.sol
dhvanipa: the entire program being empty still causes the error, it generates the table interfaces, so I think its erroring when trying to generate the system interface

dhvanipa: when I copy the exact file into `packages/world` it builds fine

Also `forge fmt` is causing diffs in codegen folder, but its ignored in `foundry.toml` so something weird is going on

package.json
```diff
@@ -7,7 +7,7 @@
     "clean": "pnpm recursive run clean",
     "deploy:dev": "pnpm --filter 'world' run dev",
     "deploy:local": "pnpm --filter 'world' deploy:local",
-    "dev": "echo clearing anvil cache... && shx rm -rf ~/.foundry/anvil/tmp && mprocs",
+    "dev": "echo clearing anvil cache... && shx rm -rf ~/.foundry/anvil/tmp && rm -rf packages/world/deploys/31337/latest.json && mprocs",
```
frolic, line 10:
```suggestion
    "dev": "echo clearing anvil cache... && shx rm -rf ~/.foundry/anvil/tmp packages/world/deploys/31337/latest.json && mprocs",
```

mprocs.yaml
```diff
@@ -2,6 +2,16 @@ procs:
...
+  default-programs:
+    cwd: packages/programs
+    shell: |
+      while [ ! -f ../world/deploys/31337/latest.json ]; do
+        echo "Waiting for world deployment..."
+        sleep 2
+      done
+      worldAddress=$(awk -v id="31337" -F'"' '$2 == id {getline; print $4}' ../world/worlds.json)
+      echo "Using world address: $worldAddress"
+      pnpm mud deploy --worldAddress=$worldAddress --alwaysRunPostDeploy=true --rpcBatch=true
```
frolic, line 14: we could prob simplify this further with something like
```suggestion
      wait-on ../worlds/deploys/31337/latest.json && pnpm mud deploy --worldAddress=$(cat ../worlds/worlds.json | jq -r ".[\"31337\"].address // empty") --alwaysRunPostDeploy=true --rpcBatch=true
```

frolic: you'll need to `pnpm add -Dw wait-on`

packages/programs/src/programs/BedProgram.sol
```diff
@@ -0,0 +1,22 @@
...
+
+contract BedProgram is DefaultProgram, ISleepHook, IWakeupHook {
+  constructor(IBaseWorld _world) DefaultProgram(_world) { }
+
+  function onSleep(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
+    require(_isAllowed(target, caller), "Only approved callers can sleep in the bed");
+  }
+
+  function onWakeup(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
+    require(_isAllowed(target, caller), "Only approved callers can wake up from the bed");
```
alvrs, line 20: kinda funny, is there anything you could do if someone changes the program to prevent you from waking up while you're asleep?

alvrs: you couldn't even wait till you're out of energy because the bed keeps your energy level lol

packages/programs/src/programs/DefaultProgram.sol
```diff
@@ -0,0 +1,55 @@
...
+  constructor(IBaseWorld _world) WorldConsumer(_world) { }
+
+  function onAttachProgram(EntityId caller, EntityId target, bytes memory extraData) external onlyWorld {
+    uint256 uniqueEntity = UniqueEntity.get() + 1;
+    UniqueEntity.set(uniqueEntity);
+    bytes32 smartItemId = bytes32(uniqueEntity);
+    SmartItem.set(target, smartItemId);
+
+    Owner.set(smartItemId, caller);
+    AllowedCaller.set(smartItemId, caller, true);
```
alvrs, line 27: why do we need to create a new `smartItemId` here instead of just using the `target` entity id?

dhvanipa: See thread above: https://github.com/dustproject/dust/pull/184#discussion_r2032028924

But tl;dr: 
1) User A places down a chest and get entity A
2) Program adds User A to allowed caller list on attach and uses entity A as the key
3) User B mines the chest (detach is called but has a fixed amount of gas, so cleanup logic could fail)
4) User B places down a chest in the same spot and gets entity A (since in the physics entities are tied to positions)
5) **Error**: Now the default program will still incorrectly have User A as an allowed caller

We fix this by creating a new smartItemId, so 5) can't happen

packages/programs/src/programs/BedProgram.sol
dhvanipa: > you couldn't even wait till you're out of energy because the bed keeps your energy level lol

you'd have to wait until the force field runs out of energy yea

packages/programs/script/PostDeploy.s.sol
dhvanipa: okay found the issue, its the namespace. `default` is a reserved word in solidity and the generated system interface has the namespace as the prefix of the function name. It also doesnt seem to like dashes

dhvanipa: okay changing the namespace fixed the compilation issues but then the mud deploy still fails when registering the system, tx reverts with some error message. Maybe because these contracts need to be passed in the world address to the constructor, and the system deploy logic doesnt do that

For now not converting them to systems!

PR #187, opened 2025-04-08
format world

frolic:


packages/world/IWorld.abi.json
frolic, line 1: moved to `world/out/IWorld.sol/IWorld.abi.json` to match MUD

PR #196, opened 2025-04-09
Small cleanup of unused logic and old TODOs

vdrg:
- Removed display system/table/etc
- Added back some notifications that were removed in the Inventory refactor (had to remove some of the notification fields)
- Removed old todo comments


packages/world/src/systems/BedSystem.sol
```diff
@@ -136,4 +95,45 @@ contract BedSystem is System {
...
+
+    PlayerUtils.removePlayerFromBed(player, bed, forceField);
+
+    BedLib.transferInventory(bed, drop);
+    // TODO: Should we safecall the program?
+  }
+}
+
+// To avoid reaching bytecode size limit
+library BedLib {
```
vdrg, line 126: just moved these to the end of the file, no changes

dhvanipa: ty for the comment, makes it easier to review!

packages/world/IWorld.abi.json
dhvanipa, line 1: these ABI files shouldnt be generated now after the changes in main

dhvanipa (approved):

PR #197, opened 2025-04-09
Fix mining block bellow seed

### LINKED ISSUE #195: mining block below seed not possible
this should require `above` to be fully grown, not `mined`
https://github.com/dustproject/dust/blob/62c498c68eaf507d07be15be7216e6141cc3ab80/packages/world/src/systems/MineSystem.sol#L131-L133

vdrg:
Fixes #195


packages/world/src/systems/MineSystem.sol
```diff
@@ -93,93 +93,66 @@ contract MineSystem is System {
...
-      _requireSeedNotFullyGrown(mined);
-    } else if (mineObjectTypeId == ObjectTypes.AnyOre) {
-      mineObjectTypeId = RandomResourceLib._collapseRandomOre(mined, coord);
+    } else if (minedType == ObjectTypes.AnyOre) {
+      minedType = RandomResourceLib._collapseRandomOre(mined, coord);
     }
 
-    // First coord will be the base coord, the rest is relative schema coords
-    Vec3[] memory coords = mineObjectTypeId.getRelativeCoords(baseCoord, Orientation._get(mined));
+    MineLib._requireMinesAllowed(caller, minedType, coord, extraData);
```
dhvanipa, line 109: Are there any re-entrancy issues from calling this before the Mass table is updated?

packages/world/src/systems/MineSystem.sol
```diff
@@ -93,93 +93,66 @@ contract MineSystem is System {
...
 
-      if (mineObjectTypeId == ObjectTypes.ForceField) {
-        destroyForceField(mined);
-      }
+    // Remove flora on top of this block
+    Vec3 aboveCoord = baseCoord + vec3(0, 1, 0);
+    // If above is a seed, the entity must exist as there are not seeds in the base terrain
+    (EntityId above, ObjectTypeId aboveTypeId) = getOrCreateEntityAt(aboveCoord);
+    if (aboveTypeId.isFlora()) {
+      _removeFlora(above, aboveTypeId, aboveCoord);
```
dhvanipa, line 126: We're making it so you cant have any floating flowers?

But this doesnt cascade right?

packages/world/src/systems/MineSystem.sol
```diff
@@ -191,6 +164,45 @@ contract MineSystem is System {
...
+    EntityId mined,
+    ObjectTypeId minedType,
+    Vec3 baseCoord,
+    bytes calldata extraData
+  ) internal {
+    // Detach program if it exists
+    ProgramId program = mined.getProgram();
+    if (program.exists()) {
+      bytes memory onDetachProgram = abi.encodeCall(IDetachProgramHook.onDetachProgram, (caller, mined, extraData));
+      program.call({ gas: SAFE_PROGRAM_GAS, hook: onDetachProgram });
```
dhvanipa, line 192: We dont clear the `EntityProgram` table here?

packages/world/src/systems/MineSystem.sol
vdrg: yeah this shouldn't cascade.. but now that I think about it we do have passthrough blocks on top of each other in the terrain right? like tall grass maybe? Also, should it be possible to place a passthrough block on top of another one?

packages/world/src/systems/MineSystem.sol
vdrg: ah good catch! We weren't doing that before either

packages/world/src/systems/MineSystem.sol
dhvanipa: yea tall grass, cactus, same with vines coming down from trees

dhvanipa: >  Also, should it be possible to place a passthrough block on top of another one?
imo it should

dhvanipa: instead of doing this for all non-passthrough blocks, we could just do it for seeds

vdrg: Ah interesting, so @alvrs mentioned that the client had the concept of flora, which I assumed was assigned to all pass through blocks. Are vines also flora in the client? I think it is easy to cascade the remove flora thing but we shouldn't do it for hanging flora

packages/world/src/systems/MineSystem.sol
```diff
@@ -93,93 +93,66 @@ contract MineSystem is System {
...
-    if (mineObjectTypeId.isMachine()) {
+    if (minedType.isMachine()) {
       (EnergyData memory machineData,) = updateMachineEnergy(mined);
       require(machineData.energy == 0, "Cannot mine a machine that has energy");
-    } else if (mineObjectTypeId.isSeed()) {
-      _requireSeedNotFullyGrown(mined);
-    } else if (mineObjectTypeId == ObjectTypes.AnyOre) {
-      mineObjectTypeId = RandomResourceLib._collapseRandomOre(mined, coord);
+    } else if (minedType == ObjectTypes.AnyOre) {
+      minedType = RandomResourceLib._collapseRandomOre(mined, coord);
```
dhvanipa, line 106: hmm where do we set the mined type if the ore is not fully mined? I think right now it would compute a new type??

packages/world/src/systems/MineSystem.sol
dhvanipa: Actually `_collapseRandomOre` would get called multiple times right now because the type will remain any_ore, if mine is re-entered, but I think the type should be set and that would solve it

packages/world/src/systems/MineSystem.sol
vdrg: collapseRandomOre sets the type

packages/world/src/systems/MineSystem.sol
dhvanipa: nvm it does set it so it'd only be called once!

packages/world/src/systems/MineSystem.sol
dhvanipa: The client's concept of "flora" is to render blocks that arent cubic. Not all these blocks are passthrough though. For eg, leaves, bamboo bush, and cactus are not pass through

Vines are indeed flora

If we cascade, could I create a tower of flowers and then mining the bottom flower would revert cuz there's too much gas used?

alvrs: I think only doing it for passthrough (so ie not for cactus) and only doing it for the one right above (or maybe up to depth 2 to catch tall grass) is fine. For vines we actually would not want to cascade all of it, since they're usually hanging down from something. Think it's fine if there are air blocks/flora in some cases, just feels we should handle the default case of wheat/grass/flower on top of a block.

alvrs: could also just say "if the top block was tall grass, remove one more" to avoid always having to check two blocks

alvrs: > But I could place a tall grass and then a solid block above, and the solid block would incorrectly be mined then

i mean only attempt mining the block above that if it is also tall grass haha

> also how do force field permission checks work here? What if the flower is in a different force field from the grass block below? You might have permission to mine the grass block but not the flower above

hm yeah that's almost a game design question, could argue that it should still be mined because it dependend on the block below it, or could argue that it should be skipped because it's protected by the force field. How does this work for beds (or other things that are larger than a single block)? Would tend towards removing it even though it's in a force field.

dhvanipa: > hm yeah that's almost a game design question

yea deff, I think either could make sense!

For beds (and other multi voxel objects), you can mine any of the individual voxels. If you mine the voxel inside the force field --> force field is called, but if you mine the voxel outside, force field is not called. So its your responsibility as a player to ensure when you place a bed that its fully inside, otherwise it can be mined

alvrs: personally i would go with:
- if block above the mined block is passthrough: destroy it, no matter what
- if the block above the mined block is tall grass, also check the one above that and destroy it if it's also tall grass

alternatively i'd also be fine with:
- only destroy the block above if it's a seed (because in that case it already depends on the farmland)

dhvanipa: > if the block above the mined block is tall grass, also check the one above that and destroy it if it's also tall grass

what should happen if I have 3 blocks stacked of tall grass? the very top one would remain floating?

alvrs: yeah exactly

vdrg: reverted to just doing it for seeds for now

packages/world/src/systems/MineSystem.sol
vdrg: I moved it to the end just to be sure

packages/world/src/ObjectTypeLib.sol
```diff
@@ -118,6 +119,10 @@ library ObjectTypeLib {
     return self == ObjectTypes.AnyLog || self == ObjectTypes.AnyPlanks;
   }
 
+  function isFlora(ObjectTypeId self) internal view returns (bool) {
```
dhvanipa, line 122: Could we remove this? It'll be misleading cuz not all flora are pass through eg, leaves

dhvanipa (approved):

PR #198, opened 2025-04-09
Craft fuel

### LINKED ISSUE #192: add fuel system
instead of crafting fuel via a recipe, allow crafting a variable amount of fuel by specifying number of logs & leaves

require powerstone

dhvanipa:
Fixes #192


packages/world/src/systems/CraftFuelSystem.sol
```diff
@@ -0,0 +1,52 @@
...
+import { ObjectTypeId } from "../ObjectTypeId.sol";
+import { ObjectTypeLib } from "../ObjectTypeLib.sol";
+import { ObjectTypes } from "../ObjectTypes.sol";
+
+import { Vec3 } from "../Vec3.sol";
+import { transferEnergyToPool } from "../utils/EnergyUtils.sol";
+import { InventoryUtils, SlotAmount } from "../utils/InventoryUtils.sol";
+import { CraftNotification, notify } from "../utils/NotifUtils.sol";
+
+contract CraftFuelSystem is System {
```
dhvanipa, line 22: Tried putting this in `CraftSystem` and combining `craftFuel` and `fuelMachine` into one system but in both cases contract too big

packages/world/src/systems/CraftFuelSystem.sol
```diff
@@ -0,0 +1,52 @@
...
+    uint128 totalEnergy = 0;
+    for (uint256 i = 0; i < inputs.length; i++) {
+      ObjectTypeId inputType = InventorySlot._getObjectType(caller, inputs[i].slot);
+      require(inputType.isLog() || inputType.isLeaf(), "Can only use logs or leaves to make fuel");
+      // we convert the mass to energy
+      totalEnergy +=
+        inputs[i].amount * (ObjectTypeMetadata._getEnergy(inputType) + ObjectTypeMetadata._getMass(inputType));
+      InventoryUtils.removeObjectFromSlot(caller, inputType, inputs[i].amount, inputs[i].slot);
+    }
+    uint128 fuelAmount = totalEnergy / ObjectTypeMetadata._getEnergy(ObjectTypes.Fuel);
```
vdrg, line 43: I wonder if we should just do fuelAmount == totalEnergy, we could have rounding errors if not. Unless we know that the energy + mass of all logs and leaves is always a multiple of fuel

packages/world/src/systems/CraftFuelSystem.sol
```diff
@@ -0,0 +1,52 @@
...
+    require(fuelAmount > 0 && fuelAmount <= uint128(type(uint16).max), "Invalid fuel amount");
+    InventoryUtils.addObject(caller, ObjectTypes.Fuel, uint16(fuelAmount));
```
vdrg, line 45: we could lift the uint16 restriction I think

packages/world/src/systems/CraftFuelSystem.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    transferEnergyToPool(caller, CRAFT_ENERGY_COST);
+
+    // TODO: should we use a diff notification for fuel?
```
vdrg, line 49: I think we should, but it is still unclear to me if we need notifications rn

vdrg (approved):
lgtm! maybe we could optimize getting the object types later by using a fixed size array

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: the energy of fuel is set to 1 right now so it should be a multiple

I can remove the division though and save a read

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: hmm cuz the cast will cause an overflow anyways?

dhvanipa:
> we could optimize getting the object types later by using a fixed size array

do you mean the return type of `getLogObjectTypes`? I thought we cant because its used in `getObjectTypes`

packages/world/src/systems/CraftFuelSystem.sol
vdrg: oh, but in that case won't fuel be too many to fit in the inventory? oak log has 12500000000000000 mass

packages/world/src/systems/CraftFuelSystem.sol
vdrg: no sorry I meant lifting it in the inventory utils and handle larger amoutns there

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: will be needed for quests, and also helpful for analytics

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: haha I mean its 1 without scaling, so fuel = 100000000000000 mass

1 oak log will give you 178 (125 mass + 53 energy) fuel

vdrg:
oh you are right, I wonder if we could remove the getObjectTypes if we use slots for crafting

packages/world/src/systems/CraftFuelSystem.sol
vdrg: ah nice ok  👍

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: lifted out to inventory utils!

packages/world/src/systems/CraftFuelSystem.sol
dhvanipa: hardcoded the energy value with a comment

dhvanipa:
> oh you are right, I wonder if we could remove the getObjectTypes if we use slots for crafting

yea I think we can when we update crafting!

PR #201, opened 2025-04-10
Fix fragment drain rate

### LINKED ISSUE #159: decrease forcefield drain rate on contraction if the fragment contains beds with players

vdrg:
- Refactored data schema to be consistent with other entities (map entityId to data instead of position)
- Encapsulated forcefield utils in a library
- Fixed bugs related to correctly setting the drain rate of fragments by storing an "extraDrainRate" field for fragments, which gets added/removed from forcefield's drain rate when adding/removing a fragment

Fixes #159


packages/world/src/utils/EnergyUtils.sol
```diff
@@ -3,14 +3,16 @@ pragma solidity >=0.8.24;
 
 import { BedPlayer } from "../codegen/tables/BedPlayer.sol";
 import { Energy, EnergyData } from "../codegen/tables/Energy.sol";
+import { console } from "forge-std/console.sol";
```
frolic, line 6: fyi

vdrg: whoops thanks

packages/world/mud.config.ts
```diff
@@ -310,6 +309,15 @@ export default defineWorld({
       },
       key: ["entityId"],
     },
+    ReverseFragmentPosition: {
```
dhvanipa, line 312: Doesnt this indirection increase gas?

packages/world/src/systems/BedSystem.sol
```diff
@@ -39,23 +46,22 @@ contract BedSystem is System {
   function sleep(EntityId caller, EntityId bed, bytes calldata extraData) public {
     caller.activate();
 
-    (Vec3 callerCoord, Vec3 bedCoord) = caller.requireConnected(bed);
+    (Vec3 callerCoord,) = caller.requireConnected(bed);
```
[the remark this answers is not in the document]
vdrg: because we need the base entity

packages/world/src/utils/EnergyUtils.sol
```diff
@@ -26,29 +27,23 @@ using ObjectTypeLib for ObjectTypeId;
...
   }
 
-  // Calculate energy drain
   uint128 energyDrained = timeSinceLastUpdate * energyData.drainRate;
-
   uint128 depletedTime = 0;
-  // Update accumulated depleted time if it ran out of energy on this update
+
   if (energyDrained >= energyData.energy) {
-    // Calculate when it ran out by determining how much time it took to drain the energy
```
dhvanipa, line 49: why were these comments removed?

packages/world/mud.config.ts
vdrg: it does, but with the new changes it felt more natural like this and more consistent, as we usually get the forcefield and the fragment entities with `getForceField`. But yeah we could try abstracting it away and using less tables, but also felt weird because for all other entities we always map entity -> data, and coord -> entity.

packages/world/src/utils/EnergyUtils.sol
vdrg: was trying to clean up some of the comments as they were kind of redundant, but yeah this one can stay

dhvanipa:
> Fixed bugs related to correctly setting the drain rate of fragments by storing an "extraDrainRate" field for fragments, which gets added/removed from forcefield's drain rate when adding/removing a fragment

So if I understand correctly now: sleeping/waking up on beds updates both the force field drain rate AND fragment drain rate, so that when a fragment is removed, all the drain rate from the beds inside is auto removed by subtracting `extraDrainRate`. That's why we need to store it, correct?

packages/world/mud.config.ts
dhvanipa: ooc what's the gas increase? If you compare gas report on main and this branch? If its not that much then yea for consistency can keep the extra table

packages/world/src/systems/BedSystem.sol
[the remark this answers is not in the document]
dhvanipa: ahh I see

dhvanipa (approved):

vdrg:
> So if I understand correctly now: sleeping/waking up on beds updates both the force field drain rate AND fragment drain rate, so that when a fragment is removed, all the drain rate from the beds inside is auto removed by subtracting extraDrainRate. That's why we need to store it, correct?

yep exactly that's the idea, and can also reuse if we ever have other entities that increase drain rates

packages/world/mud.config.ts
vdrg: on it

PR #203, opened 2025-04-14
Add remaining seeds/saplings

### LINKED ISSUE #156: spawn with half energy

### LINKED ISSUE #200: add remaining tree seeds & set energy values
the energy values need to be set to the biggest tree that can be possibly generated

vdrg:
Fixes #156 
Fixes #200


packages/world/src/Constants.sol
```diff
@@ -37,9 +37,18 @@ uint128 constant MOVE_ENERGY_COST = 25543750000000;
...
+uint256 constant MAX_PUMPKIN_SEED = 1_717;
+
+uint256 constant MAX_OAK_SEED = 44_745;
+uint256 constant MAX_BIRCH_SEED = 35_656;
+uint256 constant MAX_JUNGLE_SEED = 27_671;
+uint256 constant MAX_SAKURA_SEED = 3_437;
+uint256 constant MAX_ACACIA_SEED = 20_701;
+uint256 constant MAX_SPRUCE_SEED = 19_594;
+uint256 constant MAX_DARK_OAK_SEED = 13_732;
+uint256 constant MAX_MANGROVE_SEED = 21_610;
```
alvrs, line 51: nit: if we rename the object types to sapling, should we call these `_SAPLING` too?

alvrs: (or keep the seed name for the object types, no strong opinion)

vdrg: yep good catch, I think it might be better to use sapling, as they work different from seeds

packages/world/src/NatureLib.sol
```diff
@@ -91,6 +155,30 @@ library NatureLib {
...
   }
 
+  function getLeafDrops(ObjectTypeId objectType, uint256 randomSeed)
+    internal
+    view
+    returns (ObjectAmount[] memory result)
+  {
+    uint256[] memory distribution = new uint256[](2);
+    distribution[0] = 43; // No sapling: 43%
+    distribution[1] = 57; // 1 sapling: 57%
```
alvrs, line 165: where do these numbers come from? if we want ~1 sapling per tree shouldn't the probability for sapling be much lower?

vdrg: wasn't sure about those, they get adjusted with the current supply of seeds (so 57% is the max). So should we have different distributions for different seeds depending on the max number of leaves for each one?

alvrs: If it's not too complex to do i'd say so yeah. Would probably also turn the max drop rate way down (closer to maybe like 2-3 saplings per tree) so that saplings are not too abundant even in the beginning and there is an incentive to mine all the leaves (to avoid floating leaves everywhere)

packages/world/package.json
```diff
@@ -42,6 +42,7 @@
     "forge-std": "https://github.com/foundry-rs/forge-std.git#3b20d60d14b343ee4f908cb8079495c07f5e8981",
     "solhint": "^3.4.1",
     "solhint-config-mud": "2.2.22-ab837ceb49fa77cc29487bb9df0c487975b37afe",
-    "solhint-plugin-mud": "2.2.22-ab837ceb49fa77cc29487bb9df0c487975b37afe"
+    "solhint-plugin-mud": "2.2.22-ab837ceb49fa77cc29487bb9df0c487975b37afe",
+    "typescript": "^5.8.3"
```
dhvanipa, line 46: why do we need this here if typescript is already installed in the root package.json?

packages/world/src/Constants.sol
```diff
@@ -37,9 +37,18 @@ uint128 constant MOVE_ENERGY_COST = 25543750000000;
 uint128 constant PLAYER_FALL_ENERGY_COST = MAX_PLAYER_ENERGY / 25; // This makes it so, with full energy, you die from a 25 block fall
 
 // Resource caps
-uint256 constant MAX_WHEAT_SEED = 5_000_000;
-uint256 constant MAX_OAK_SEED = 1_000_000;
-uint256 constant MAX_SPRUCE_SEED = 1_000_000;
+uint256 constant MAX_WHEAT_SEED = 444_971;
```
dhvanipa, line 40: We should update the new energy buffer value in this PR as well

packages/world/src/systems/BuildSystem.sol
```diff
@@ -57,7 +55,7 @@ contract BuildSystem is System {
 
     (EntityId base, Vec3[] memory coords) = BuildLib._addBlocks(coord, buildObjectType, direction);
 
-    if (buildObjectType.isSeed()) {
+    if (buildObjectType.isGrowable()) {
       BuildLib._handleSeed(base, buildObjectType, coord);
```
dhvanipa, line 59: nit: should rename handleSeed to handleGrowable

packages/world/src/systems/MineSystem.sol
```diff
@@ -153,7 +150,7 @@ contract MineSystem is System {
 
   function _removeBlock(EntityId entityId, ObjectTypeId objectType, Vec3 coord) internal {
     // If object being mined is seed, no need to check above entities
-    if (objectType.isSeed()) {
+    if (objectType.isGrowable()) {
       _removeSeed(entityId, objectType, coord);
```
dhvanipa, line 154: nite: rename to removeGrowable

packages/world/src/systems/FarmingSystem.sol
```diff
@@ -35,150 +33,8 @@ contract FarmingSystem is System {
...
-    }
-  }
-}
-
-library FarmingLib {
-  function _processEnergyReduction(EntityId caller) public {
     transferEnergyToPool(caller, TILL_ENERGY_COST);
-  }
-
-  function _growTree(EntityId seed, Vec3 baseCoord, TreeData memory treeData) public returns (uint32, uint32) {
```
dhvanipa, line 93: Was a bunch of this code modified or just moved to NatureSystem?

vdrg: Yeah a lot changed, now we operate over the predefined tree data instead of doing calculations at runtime

packages/world/src/systems/NatureSystem.sol
```diff
@@ -33,40 +45,162 @@ contract NatureSystem is System {
...
+
+    for (uint256 j = 0; j < randomLeaves.length; ++j) {
+      Vec3 rel = randomLeaves[j];
+      if (rel.y() > int32(trunkHeight)) {
+        break;
+      }
+
+      rand = uint256(keccak256(abi.encodePacked(rand, j))); // evolve RNG
+
+      if (rand % 100 < 40) continue; // 40 % trimmed
```
dhvanipa, line 169: what is this work?

packages/world/src/TreeLib.sol
```diff
@@ -0,0 +1,136 @@
+// SPDX-License-Identifier: MIT
+/* Auto‑generated. DO NOT EDIT. */
+pragma solidity >=0.8.24;
+
+import { ObjectTypeId } from "./ObjectTypeId.sol";
+import { ObjectTypes } from "./ObjectTypes.sol";
+import { Vec3 } from "./Vec3.sol";
+
+library TreeBlobs {
+  /* Oak – 50 fixed, 8 random */
```
dhvanipa, line 10: fixed == all oak trees will have 50 leaves
random = an oak tree can have 0-8 additional leaves

oak tree will have 50-58 leaves

is that correct?

dhvanipa (approved):

packages/world/src/systems/NatureSystem.sol
vdrg: the original tree algo applied 40% chance of creating a leaf for certain leafs like corners

packages/world/src/TreeLib.sol
vdrg: yup

packages/world/src/Constants.sol
vdrg: could you remind me how we calculated that before?

packages/world/package.json
vdrg: because it would fail to build:js unless you have it installed globally

packages/world/src/Constants.sol
dhvanipa: should just be `Energy Buffer in each Energy Region` cell times the factor

dhvanipa: also looks like `max player energy bar` changed

dhvanipa: actually a bunch changed, could do in follow up PR lol nvm

PR #206, opened 2025-04-15
Remaining Inventory issues

### LINKED ISSUE #202: support transferring within same inventory
- moving items around in player inventory
- moving items around in chest inventory

### LINKED ISSUE #208: check object type on attachProgram
only allow smart items to have programs

vdrg:
Fixes #202 
Fixes #208 

-This should now support self transfers for players and chests.
- Also sneaked in the change to only allow programs for smart entities.
- Adapts remaining systems to use slots.


packages/world/src/systems/TransferSystem.sol
```diff
@@ -24,56 +23,46 @@ contract TransferSystem is System {
...
-    EntityId target = _getTarget(caller, from, to);
-
-    require(ObjectType._get(target) != ObjectTypes.Player, "Cannot transfer to player");
-
-    bytes memory onTransfer =
-      abi.encodeCall(ITransferHook.onTransfer, (caller, target, from, to, objectAmounts, entities, extraData));
+    // Transferring within the same inventory
+    if (from == to) {
+      if (caller != from) {
+        caller.requireConnected(from);
```
dhvanipa, line 36: Do we need a player check here as well? What if I call transfer with: caller = player A, from = player B, to = player B

Ie moving objects around in another players inventory

dhvanipa (commented):
Merging this in will break client so should wait until a corresponding branch updates calling build & craft

packages/world/src/systems/TransferSystem.sol
vdrg: ah lmao you are right

dhvanipa: would be good to add a test for this case as well

vdrg: yeah we need a test sprint for sure, rn I'm just trying to get everything to a working state

packages/world/src/ProgramInterfaces.sol
```diff
@@ -35,10 +36,8 @@ interface ITransferHook {
   function onTransfer(
     EntityId caller,
     EntityId target,
-    EntityId from,
-    EntityId to,
-    ObjectAmount[] memory objectAmounts,
-    EntityId[] memory entities,
+    SlotData[] memory deposits,
```
dhvanipa, line 39: What does this look like for a transfer within the same chest? ie moving items around

As a chest program writer, is that a "depsoit"/"withdraw"? or I guess both would be filled with the same items?

vdrg: if you move stuff within the chest's inventory, let's say the object in slot 0 to slot 1, you would see deposits[0] == withdrawals[0].

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -257,8 +242,21 @@ library InventoryUtils {
 
       InventorySlotData memory sourceSlot = InventorySlot._get(from, slotFrom);
       require(!sourceSlot.objectType.isNull(), "Empty slot");
+      fromSlotData[i] = SlotData(sourceSlot.entityId, sourceSlot.objectType, sourceSlot.amount);
 
       InventorySlotData memory destSlot = InventorySlot._get(to, slotTo);
+
+      // Handle slot swaps (transferring all to an existing slot with a different type)
```
dhvanipa, line 249: What happens right now if you try to combine two slots? one slot with 30 dirt, another with 10 dirt. Would it become one slot with 40 dirt? It wouldnt swap I think cuz you're checking if they're diff object types

vdrg: it would just follow the regular transfer flow from one slot to another, but will add a test

dhvanipa (approved):

alvrs:
client integration: https://github.com/dustproject/dust-client/pull/698

PR #212, opened 2025-04-16
add any leaf

dhvanipa:


packages/world/script/initObjects.sol
```diff
@@ -1709,4 +1709,14 @@ function initObjects() {
       canPassThrough: false
     })
   );
+  ObjectTypeMetadata.set(
+    ObjectTypes.AnyLeaf,
+    ObjectTypeMetadataData({
+      stackable: 99,
```
vdrg, line 1715: do Any objects need to be stackable, have mass, etc?

PR #216, opened 2025-04-17
fix empty bytes32

dhvanipa:


packages/dustkit/src/apps/getProgramAppConfigUrl.ts
```diff
@@ -26,7 +26,10 @@ export async function getProgramAppConfigUrl({
       systemId: program,
       abi: appAbi,
       functionName: "appConfigURI",
-      args: [entity ?? "0x"],
+      args: [
+        entity ??
+          "0x0000000000000000000000000000000000000000000000000000000000000000",
```
frolic, line 31: oops, previous comment was in a commit rather than PR

could use `zeroHash` from viem for this

frolic (approved):

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
dhvanipa: updated! Didnt know this existed, can replace the same string in the client too

frolic: there's also `zeroAddress`

PR #217, opened 2025-04-17
dont throw if ContractFunctionRevertedError

dhvanipa:


packages/dustkit/src/apps/getProgramAppConfigUrl.ts
```diff
@@ -18,17 +18,30 @@ export async function getProgramAppConfigUrl({
...
+    // systems so that their hook functions can only be called by the world
+    url = await readContract(client, {
+      address: system,
       abi: appAbi,
       functionName: "appConfigURI",
       args: [entity ?? zeroHash],
-    }),
-  });
+    });
+  } catch (error: any) {
```
frolic, line 39: there's examples in mud for how to do this better than with any

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
```diff
@@ -18,17 +18,30 @@ export async function getProgramAppConfigUrl({
...
+    if (error?.name === "ContractFunctionExecutionError") {
```
frolic, line 40: can we be any more specific than this?

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
```diff
@@ -18,17 +18,30 @@ export async function getProgramAppConfigUrl({
   program: ProgramId;
   entity?: EntityId;
 }): Promise<string | undefined> {
-  const url = await readContract(client, {
+  let url: string | undefined;
```
frolic, line 21: you can move this and the empty string check inside the try/catch and then you don't need a let hanging around

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
dhvanipa: updated

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
dhvanipa: updated

packages/dustkit/src/apps/getProgramAppConfigUrl.ts
dhvanipa: image

dhvanipa: This is the error I get when theres no method

image

PR #221, opened 2025-04-18
Kill when not enough energy for an action

### LINKED ISSUE #193: taking an action that goes over the player's energy should kill the player, not revert with `Not enough energy`
The current behavior (reverting instead of depleating the energy) leads to weird situations where players are stuck not being able to do actions because of the `Not enough energy` revert, but also not being able to respawn because they still have some small amount of energy left.

vdrg:
Fixes #193


packages/world/src/systems/MineSystem.sol
```diff
@@ -109,7 +109,17 @@ contract MineSystem is System {
...
 
-    uint128 finalMass = MassReductionLib._processMassReduction(caller, callerCoord, mined, toolSlot);
+    (uint128 finalMass, uint128 energyReduction) = _processEnergyReduction(caller, callerCoord, mined, toolSlot);
+
+    {
+      if (energyReduction > 0) {
+        // If player died, return early
+        (uint128 callerEnergy,) = transferEnergyToPool(caller, energyReduction);
+        if (callerEnergy == 0) {
+          return mined;
```
dhvanipa, line 119: hmm this will consume tool mass but not set the terrain mass right?

vdrg: ah you are right, but it is kind of a chicken and egg problem 🤔  because we need to use the tool to see if we need to consume from player

packages/world/src/systems/MineSystem.sol
```diff
@@ -217,6 +221,52 @@ contract MineSystem is System {
...
+
+    // if tool mass reduction is not enough, consume energy from player up to mine energy cost
+    uint128 energyReduction = 0;
+    if (toolMassReduction < massLeft) {
+      uint128 remaining = massLeft - toolMassReduction;
+      energyReduction = MINE_ENERGY_COST <= remaining ? MINE_ENERGY_COST : remaining;
+      massLeft -= energyReduction;
+    }
+
+    uint128 finalMass = massLeft - toolMassReduction;
```
dhvanipa, line 266: can this underflow?

vdrg: no, because we pass massLeft to `getMassReduction`, which limits the amount of mass that can be reduced

dhvanipa (commented):
in both mine and hit, the player will lose energy that doesnt convert into mass being lost in the block or energy being lost in the machine if they die, but that should be fine cuz we just transfer it to the local pool, so things are still conserved

dhvanipa (approved):

packages/world/src/systems/MineSystem.sol
dhvanipa: ahh and then also when we subtract from massLeft in the player case we're using remaining, nice

vdrg: also in lines 259-264, we only reduce massLeft up to `massLeft - toolMassReduction`, so it can never be smaller than `toolMassReduction`

PR #223, opened 2025-04-21
increase placeable sizes

dhvanipa:


dhvanipa:
client having trouble rendering https://github.com/dustproject/dust-client/pull/723

Issue #40, opened 2025-02-14
Refactor ObjectMetadata table


vdrg:
- Extract `canPassThrough` and `maxInventorySlots`
- Could even extract `stackable` as it doesn't apply to all objects
- ObjectMetadata could become ObjectPhysicsData (only mass and energy)

vdrg:
I believe maxInventory slots can be obtained through a function, as there are only 3 objects that have inventory. Maybe when refactoring the ObjectTypes to include subcategories we could just make ObjectTypeId an uint256, and encode all this data in there to completely remove the table

PR #224, opened 2025-04-21
Refactor object types

vdrg:
- [x] Add codegen for object types
 - Use efficient category checks
 - Use efficient meta category mask checks
- [x] Remove fields from ObjectTypeMetadata table
- [x] Rename ObjectType table to EntityObjectType
- [x] Add missing ObjectTypeLib functions  
- [x] Review that recipes and mass/energy matches original
- [ ] Fix recipes that use coal


packages/world/mud.config.ts
```diff
@@ -77,19 +77,16 @@ export default defineWorld({
     // ------------------------------------------------------------
     ObjectTypeMetadata: {
```
dhvanipa, line 78: should we rename this to `ObjectMassEnergy` or something other than metadata so that we can use that for later

packages/world/package.json
```diff
@@ -6,21 +6,26 @@
...
   },
-  "files": ["src", "out", "dist", "worlds.json"],
+  "files": ["src", "out", "dist", "ts", "worlds.json"],
   "scripts": {
-    "build": "pnpm run build:mud && pnpm run build:js",
+    "build": "pnpm run codegen && pnpm run build:mud && pnpm run build:js",
     "build:js": "tsc",
     "build:mud": "mud build && abi-ts --extension .ts",
     "clean": "forge clean && shx rm -rf src/**/codegen",
+    "codegen": "pnpm run codegen:object-type && pnpm run codegen:init-objects && pnpm run codegen:init-recipes && pnpm run codegen:tree-lib && forge fmt",
```
dhvanipa, line 24: nit: could we codegen just one file that does all of these?

packages/world/src/NatureLib.sol
```diff
@@ -196,25 +187,26 @@ library NatureLib {
...
+    ObjectAmount[] memory oreOptions = new ObjectAmount[](6);
+    uint256[] memory weights = new uint256[](6);
 
-    ObjectTypeId[] memory oreTypes = getOreObjectTypes();
+    ObjectType[7] memory oreTypes = ObjectTypeLib.getOreTypes();
     // Use remaining amounts directly as weights
-    for (uint256 i = 0; i < weights.length; i++) {
-      oreOptions[i] = ObjectAmount(oreTypes[i], 1);
-      (, weights[i]) = getCapAndRemaining(oreTypes[i]);
+    // Skip AnyOre (index 0) since it's not a specific ore type
```
dhvanipa, line 199: why include anyOre in `getOreTypes`?

packages/world/src/ObjectType.sol
```diff
@@ -0,0 +1,925 @@
...
+
+  function getMaxInventorySlots(ObjectType self) internal pure returns (uint16) {
+    if (self == ObjectTypes.Player) return 36;
+    if (self == ObjectTypes.Chest) return 27;
+    if (self.isPassThrough()) return type(uint16).max;
+    return 0;
+  }
+
+  function getStackable(ObjectType self) internal pure returns (uint16) {
+    if (self.isUniqueObject()) return 1;
```
dhvanipa, line 786: I think before stackable being 0 made sure that you couldnt add invalid objects to your inventory (eg water/fragment), is this still the case?

packages/world/src/ObjectType.sol
```diff
@@ -0,0 +1,925 @@
...
+      ObjectTypes.Blackstone,
+      ObjectTypes.CobbledDeepslate
+    ];
+  }
+
+  function getGemstoneTypes() internal pure returns (ObjectType[2] memory) {
+    return [ObjectTypes.Amethyst, ObjectTypes.Glowstone];
+  }
+
+  function getSoilTypes() internal pure returns (ObjectType[9] memory) {
```
dhvanipa, line 454: This lib is quite big, does forge only include functions that a system uses or all of this bytecode?

packages/world/src/systems/helper/ReadSystem.sol
dhvanipa, line 1: Can do in a separate PR but `ReadSystem` can be removed now since the client doesnt use getters

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -121,9 +119,9 @@ library InventoryUtils {
     return slot;
   }
 
-  function addObject(EntityId owner, ObjectTypeId objectType, uint128 amount) public {
+  function addObject(EntityId owner, ObjectType objectType, uint128 amount) public {
     require(amount > 0, "Amount must be greater than 0");
-    uint16 stackable = ObjectTypeMetadata._getStackable(objectType);
+    uint16 stackable = objectType.getStackable();
     require(stackable > 0, "Object type cannot be added to inventory");
```
dhvanipa, line 125: re previous comment about stackable, this would never get triggered right

packages/world/ts/internal/genInitRecipesSol.ts
```diff
@@ -0,0 +1,52 @@
+import type { ObjectTypeName } from "../objects";
+import { type Recipe, recipes } from "../recipes";
+
+function renderRecipe(recipe: Recipe): string {
```
dhvanipa, line 4: We should do the check here to ensure that the mass & energy of outputs is as expected, ie conserved

packages/world/ts/objects.ts
```diff
@@ -0,0 +1,520 @@
...
+    // TODO: must assign some of these to terrain categories
+    { name: "RedMushroomBlock", mass: 12500000000000000n },
+    { name: "BrownMushroomBlock", mass: 12500000000000000n },
+    { name: "MushroomStem", mass: 12500000000000000n },
+    { name: "Cactus", mass: 1300000000000000n },
+    { name: "Snow", mass: 4000000000000000n },
+    { name: "Ice", mass: 4000000000000000n },
+    { name: "SpiderWeb", mass: 100000000000000n },
+    { name: "Bone", mass: 1000000000000000n },
+    { name: "TextSign", mass: 17800000000000000n },
```
dhvanipa, line 398: TextSign should be in SmartEntityBlock

packages/world/ts/objects.ts
```diff
@@ -0,0 +1,520 @@
...
+// All categories metadata for template generation
+export const allCategoryMetadata: CategoryMetadata[] = [
+  ...blockCategoryMetadata,
+  ...nonBlockCategoryMetadata,
+];
+
+// Meta-categories (categories that should be included in pass-through check)
+export const passThroughCategories: Category[] = [
+  "NonSolid",
+  "Leaf",
```
dhvanipa, line 80: All the leaves, bamboo_bush, and cactus are not pass through

packages/world/ts/internal/genInitRecipesSol.ts
dhvanipa: ahh I see `validateRecipeMassEnergy` but doesnt seem like its getting called anywhere?

dhvanipa:
We'll need to update the client before we merge this in:

- [ ] update it to use new codegen files
- [ ] update scripts that set mass and energy
- [ ] update sync code for `ObjectType` table

dhvanipa:
Fixes https://github.com/dustproject/dust/issues/40

packages/world/package.json
vdrg: you mean having a bash script for this? It is kind of the same as just doing it here no?

packages/world/src/NatureLib.sol
vdrg: it is just the automatic getter generated for all categories

packages/world/src/ObjectType.sol
vdrg: yes, internal functions are inlined so it can be as big as we need

packages/world/src/utils/InventoryUtils.sol
vdrg: good point!

packages/world/ts/objects.ts
vdrg: is it? I thought we were not going to use programs for signs

packages/world/ts/objects.ts
vdrg: ah true true

packages/world/src/NatureLib.sol
vdrg: if we follow the convention of placing the AnyX type at the beginning of the category we get those cheap checks I wrote

packages/world/package.json
dhvanipa: yea cuz its easier to read than a bunch of chained `&&`, but not a strong preference, can keep what you have

packages/world/ts/objects.ts
dhvanipa: ahh no we're not going to support content of signs in the physics, which means all content has to come from programs right

packages/world/ts/internal/genInitRecipesSol.ts
vdrg: yeah wasn't using it, now I'm validating on gen, but currently only emitting a warning due to the furnace recipes (they have coal input which shouldn't be counted as part of the recipe)

packages/world/src/systems/FarmingSystem.sol
```diff
@@ -4,43 +4,41 @@ pragma solidity >=0.8.24;
...
-  using ObjectTypeLib for ObjectTypeId;
-
   function till(EntityId caller, Vec3 coord, uint16 toolSlot) external {
     caller.activate();
     (Vec3 callerCoord,) = caller.requireConnected(coord);
 
-    (EntityId farmland, ObjectTypeId objectTypeId) = getOrCreateEntityAt(coord);
-    require(objectTypeId == ObjectTypes.Dirt || objectTypeId == ObjectTypes.Grass, "Not dirt or grass");
+    (EntityId farmland, ObjectType objectType) = getOrCreateEntityAt(coord);
+    require(objectType == ObjectTypes.Dirt || objectType == ObjectTypes.Grass, "Not dirt or grass");
```
dhvanipa, line 30: @vdrg thoughts on also putting this into `ObjectType`? something like `isTillable`?

vdrg: Yeah makes sense to me

PR #227, opened 2025-04-23
fix: remove null type slots on transferAll

vdrg:


packages/world/src/utils/InventoryUtils.sol
```diff
@@ -340,6 +339,10 @@ library InventoryUtils {
       }
     }
 
+    if (InventoryTypeSlots._length(from, ObjectTypes.Null) > 0) {
```
dhvanipa, line 342: the bug was it wasnt clearing the null slot?

dhvanipa (approved):

packages/world/src/utils/InventoryUtils.sol
vdrg: yes

PR #228, opened 2025-04-25
Furnace

### LINKED ISSUE #190: add time delay coal burning mechanism for crafting
- Right now we burn coal ores when they're used in crafting
- We should make it so there's a certain amount of time you need to wait before the craft is finished
- If you mine the furnace before that time, then the furnace + the raw materials are dropped
- Only allow one thing to be crafted at a time

vdrg:
- Furnace has 2 slots (currently all furnace recipes are only coal and another ore)
- Flow:
  - Deposit coal and ore to furnace (doesn't matter which slot, but client can define specific ones)
  - Call craftWithStation, passing in the input slots of the furnace's inventory instead of the player
    - This will burn the coal immediately, and leave the ore in the furnace's inventory
    - This calls the onTransferHook on the furnace
  - After the recipe time has passed, call finishSmelting
    - This burns the ore from the furnace
    - Transfers the result to the player's inventory (maybe we want to be able to specify the output slots?)
    - Also calls the onTransferHook of the furnace

Fixes #190 

Main issue right now is the we have different categories for stations and for smart objects, not clear what the best categorization for these is


packages/world/mud.config.ts
```diff
@@ -87,6 +87,7 @@ export default defineWorld({
       schema: {
         recipeId: "bytes32",
         stationTypeId: "ObjectType",
+        smeltTime: "uint128",
```
dhvanipa, line 90: We'll need to update the client recipe ids to include smeltTime.

The furnace UI can happen after though and doesnt need to block this PR

packages/world/src/systems/CraftSystem.sol
```diff
@@ -38,19 +37,83 @@ contract CraftSystem is System {
...
-    _createRecipeOutputs(caller, recipe);
+    // If the recipe requires a furnace, the input slots reference the furnace
+    if (recipe.stationTypeId == ObjectTypes.Furnace) {
+      CraftLib._beginSmelting(station, recipeId, recipe, inputs);
+    } else {
+      CraftLib._consumeRecipeInputs(caller, recipe, inputs);
+      CraftLib._createRecipeOutputs(caller, recipe);
 
-    notify(caller, CraftNotification({ recipeId: recipeId, station: station }));
+      notify(caller, CraftNotification({ recipeId: recipeId, station: station }));
```
dhvanipa, line 50: why dont we notif on smelt?

dhvanipa: ahh nvm I see you added it when `finishSmelting` is called

packages/world/src/systems/CraftSystem.sol
```diff
@@ -70,33 +133,42 @@ contract CraftSystem is System {
         currentInput++;
       }
 
-      // Handle special case for coal ore
-      if (recipeType == ObjectTypes.CoalOre) {
+      // Handle special case for ores
+      if (recipeType.isOre()) {
```
dhvanipa, line 137: Do we need this now? Ores are not used in any other recipes

And if they are eg to craft some dense ore block we dont want to burn them, as we'll just make that recipe reversible

We dont have any at the moment though so should just remove it imo

vdrg: ah good point, yeah we should remove

packages/world/src/ObjectType.sol
```diff
@@ -714,6 +714,11 @@ library ObjectTypeLib {
   }
 
   // Specialized getters
+
+  function isStation(ObjectType self) internal pure returns (bool) {
```
dhvanipa, line 718: What are we using this for?

vdrg: not really using it, I thought it was good to have it, but can remove for simplicity

dhvanipa (commented):
If someone mines a furnace, we need to clean up any existing smelting right? I think the drops we wont need to transfer to air, as the entity id remains the same

dhvanipa (approved):

PR #232, opened 2025-04-25
add isTillable and isPlantableOn

dhvanipa:


packages/world/ts/objects.ts
```diff
@@ -314,7 +314,6 @@ export interface ObjectDefinition {
   energy?: bigint;
   timeToGrow?: bigint;
   isTillable?: boolean;
-  stackable?: number;
```
dhvanipa, line 317: removed stackable as it's not used, since we dont set it per object

packages/world/ts/internal/genObjectTypeSol.ts
```diff
@@ -276,6 +275,24 @@ ${allCategoryMetadata
...
 
+  function isTillable(ObjectType self) internal pure returns (bool) {
+    ${objects
+      .filter((obj) => obj.isTillable)
+      .map((obj) => `if (self == ObjectTypes.${obj.name}) return true;`)
+      .join("\n    ")}
+    return false;
+  }
+
+  function isPlantableOn(ObjectType self, ObjectType on) internal pure returns (bool) {
```
dhvanipa, line 286: I thought of adding this per object but currently its actually per category, so didnt add it there

vdrg (approved):

PR #244, opened 2025-05-01
Fix add entity to slot

vdrg:


packages/world/src/utils/InventoryUtils.sol
```diff
@@ -100,9 +100,37 @@ library InventoryUtils {
...
-    require(entityId.exists(), "Entity must exist");
+    ObjectType objectType = EntityObjectType._get(entityId);
+    require(!objectType.isNull(), "Entity must exist");
+
+    InventorySlotData memory slotData = InventorySlot._get(owner, slot);
+
+    require(slotData.objectType.isNull(), "Slot must be empty");
+
+    // Entities always have amount 1
+    InventorySlot._setAmount(owner, slot, 1);
```
dhvanipa, line 111: What was happening before when we didnt do all these writes?

dhvanipa (approved):

packages/world/src/utils/InventoryUtils.sol
vdrg: addEntityToSlot is only used in transfer, and it would work for most cases, except for when you do:

- add object to slot
- remove it
- transfer a tool from a differnt slot to this slot
- in this case the occupied slot would not be set

PR #246, opened 2025-05-05
fix: useEmptySlot

vdrg:
Will try to come up with a more efficient way of finding empty slots, wip

tldr of the issue:
If you did the following actions:
- Add object to inventory without specifying slot (uses slot 0)
- Add object to slot 2
- Add another object without specifying slot

This would result in duplicated occupiedIndex for slot 2, as `useEmptySlot` was incorrectly using the occupiedIndex length as the slot to add.

### DIFF
--- a/packages/world/src/utils/InventoryUtils.sol
+++ b/packages/world/src/utils/InventoryUtils.sol
@@ -378,44 +378,6 @@ library InventoryUtils {
     Inventory._deleteRecord(from);
   }
 
-  function getSlotData(EntityId owner, SlotTransfer[] memory slotTransfers)
-    internal
-    view
-    returns (EntityId[] memory entities, ObjectAmount[] memory objects)
-  {
-    // Count entities and objects
-    uint256 entityCount = 0;
-    uint256 objectCount = 0;
-
-    for (uint256 i = 0; i < slotTransfers.length; i++) {
-      SlotTransfer memory slotTransfer = slotTransfers[i];
-      if (InventorySlot._getEntityId(owner, slotTransfer.slotFrom).exists()) {
-        entityCount++;
-      } else {
-        objectCount++;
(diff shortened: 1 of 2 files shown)


packages/world/src/utils/InventoryUtils.sol
```diff
@@ -474,10 +436,18 @@ library InventoryUtils {
     }
 
     uint16 maxSlots = EntityObjectType._get(owner).getMaxInventorySlots();
-    require(occupiedIndex < maxSlots, "All slots used");
-    InventorySlot._setOccupiedIndex(owner, occupiedIndex, occupiedIndex);
-    Inventory._push(owner, occupiedIndex);
-    return occupiedIndex;
+    for (uint16 i = 0; i < maxSlots; i++) {
```
dhvanipa, line 439: I wonder if we're looping like this we could go the simpler solution of not storing occuied index. Ie that idea we had before of Inventory just being stored in a single table

dhvanipa (approved):

packages/world/src/utils/InventoryUtils.sol
vdrg: Let's think through this. My main concern is that it could become really expensive to list through inventories for entities with max slots 2^16. Like on transfer all to air on death, you could get stuck without being able to die as for each of the player slots you would need to iterate through air slots until you find an empty slot or one with enough capacity for the object type.. are you thinking of something different?

vdrg: we could make air not have an inventory, but then pickup and drop would need to change to not be slot based, which maybe is fine

vdrg: will give it a go, hopefully it doesn't change the external api too much

PR #247, opened 2025-05-05
Clean up, add more tests, efficient tracking of empty slots

vdrg:
tldr:
- Keep track of next non-tracked slot
- If objects are added, it first tries to take from tracked null slots (same as before). If null slots are empty, use the next non-tracked slot
- If an object/entity is added to a specific slot that is > next non-tracked slot, it loops through non-tracked slots and adds them to null slots
- So the only O(n) operation is when adding an object to a slot that leaves a gap between next non-tracked slot and slot (where n = slot - next <= maxSlots). And as there is no point to add stuff directly to air inventory slots, this is fine

### DIFF
--- a/packages/world/mud.config.ts
+++ b/packages/world/mud.config.ts
@@ -202,6 +202,7 @@ export default defineWorld({
     Inventory: {
       schema: {
         owner: "EntityId",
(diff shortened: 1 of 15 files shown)


dhvanipa (approved):
lgtm!

PR #249, opened 2025-05-06
add `dustClient_getPlayerPosition` DustKit API endpoint

karooolis:


packages/dustkit/src/rpc/schemas.ts
```diff
@@ -32,6 +32,19 @@ export type ClientRpcSchema = RpcSchema.From<
           }
       >;
     }
+  | {
+      Request: {
+        method: "dustClient_playerPosition";
```
dhvanipa, line 37: nit: should be `getPlayerPosition`, ie `get` prefix to clearly indicate what its doing

dhvanipa (approved):

PR #250, opened 2025-05-06
Tool multiplier

vdrg:
Ended up using specific object checks, felt more natural than just splitting the categories. Still need to add tests, might add tests for the multiplier plus some more mining tests


packages/world/src/systems/MineSystem.sol
```diff
@@ -217,50 +224,61 @@ contract MineSystem is System {
...
-      if (callerEnergy == 0) {
-        return (finalMass, false);
+  function _applyMassReduction(
+    EntityId caller,
+    Vec3 callerCoord,
+    uint16 toolSlot,
+    ObjectType minedType,
+    uint128 massLeft
+  ) internal returns (uint128, bool) {
+    if (massLeft != 0) {
```
dhvanipa, line 234: nit: can we return early instead of nesting conditionals?

packages/world/ts/objects.ts
```diff
@@ -314,10 +326,8 @@ export interface ObjectDefinition {
   mass?: bigint;
   energy?: bigint;
   timeToGrow?: bigint;
-  isTillable?: boolean;
```
dhvanipa, line 317: why was isTillable removed?

dhvanipa: ahh I see its a meta category

packages/world/src/systems/MineSystem.sol
```diff
@@ -217,50 +224,61 @@ contract MineSystem is System {
...
+        // If player died, return early
+        (uint128 callerEnergy,) = transferEnergyToPool(caller, energyReduction);
+        if (callerEnergy == 0) {
+          return (massLeft, false);
+        }
+
+        massLeft -= energyReduction;
       }
+
+      // Apply tool usage after decreasing player energy so we make sure the player is alive
```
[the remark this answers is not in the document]
vdrg: I think I did it like this originally because I saw it as "if the player does not have enough energy for an action, that action is not performed but the player still dies"

packages/world/ts/objects.ts
vdrg: I can add it back if you need it in the client

dhvanipa: no need, I can just update the client to read the metaCategory

packages/world/src/systems/MineSystem.sol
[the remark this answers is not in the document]
dhvanipa: ahh right if `canMine` is false, then we dont set `Mass._setMass`

hmm that feels a bit odd I guess, cuz the player is dying from exerting the energy + tool's mass to mine which causes them to die, vs right now they just die with no effect on the world lol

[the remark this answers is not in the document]
vdrg: imo it is just a matter of conventions, if we define that dying during an action means no changes to the world I think it's fine

packages/world/src/systems/MineSystem.sol
```diff
@@ -222,50 +232,67 @@ contract MineSystem is System {
...
+    if (massLeft == 0) {
+      return (0, true);
+    }
+
     ToolData memory toolData = InventoryUtils.getToolData(caller, toolSlot);
-    (uint128 finalMass, uint128 toolMassReduction, uint128 energyReduction) = _getMassReduction(toolData, mined);
+    uint128 toolMultiplier = _getToolMultiplier(toolData.toolType, minedType);
+
+    // round up since we want to use up to massLeft after applying the multiplier
+    uint128 toolMassReduction = toolData.getMassReduction(massLeft.divUp(toolMultiplier));
```
dhvanipa, line 250: could we not instead multiply the multiplier inside of `getMassReduction` and then you wouldnt need to divide and get rounding errors?

vdrg: did something similar in the end, did some refactoring to pass the multiplier to the tool and return both the tool and the total mass reduction. It wasn't as simple as just multiplying within the getMassReduction function because we need to figure out the tool mass reduction after applying the multiplier (basically we get the proportional amount of mass to reduce from the tool depending on how much mass we are able to reduce from massLeft)

packages/world/src/systems/MineSystem.sol
```diff
@@ -87,8 +97,8 @@ contract MineSystem is System {
   }
 
   function _mine(EntityId caller, Vec3 coord, uint16 toolSlot, bytes calldata extraData) internal returns (EntityId) {
-    caller.activate();
-    (Vec3 callerCoord,) = caller.requireConnected(coord);
+    EnergyData memory callerEnergy = caller.activate();
```
dhvanipa, line 100: nit: should collapse this to uint128 callerEnergy by adding`.energy` after `.activate()` like in the other systems

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -40,46 +40,52 @@ struct ToolData {
...
-  function useTool(EntityId owner, Vec3 ownerCoord, uint16 slot, uint128 useMassMax)
-    public
-    returns (ObjectType toolType)
-  {
+  function useTool(EntityId owner, uint16 slot, uint128 useMassMax) public returns (ObjectType toolType) {
     ToolData memory toolData = getToolData(owner, slot);
 
-    uint128 massReduction = toolData.getMassReduction(useMassMax);
-    applyMassReduction(toolData, ownerCoord, massReduction);
+    (, uint128 toolMassReduction) = toolData.getMassReduction(useMassMax, 1);
```
dhvanipa, line 61: why do we hardcode a multiplier here? or I guess where is this used? since for mines and hit we dont call this

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -40,46 +40,52 @@ struct ToolData {
...
-    return massReduction;
+    uint128 toolMass = ObjectPhysics._getMass(toolData.toolType);
+
+    // Limit to 10% of max mass per use
+    uint128 maxToolMassReduction = Math.min(toolMass / 10, toolData.massLeft);
+
+    uint128 massReduction = Math.min(maxToolMassReduction * multiplier, massLeft);
+
+    // Reverse operation to get the proportional tool mass reduction
+    uint128 toolMassReduction = massReduction / multiplier;
```
dhvanipa, line 83: hmm can there be rounding errors here? Or I guess it doesnt matter cuz this is just about picking how much mass to reduce from the tool

dhvanipa (approved):

packages/world/src/utils/InventoryUtils.sol
vdrg: yeah there will be rounding errors, but I believe it is fine because toolMassReduction will always be <= maxToolMassReduction

PR #253, opened 2025-05-07
Tests

vdrg:
- Add more test for multiple systems
- Fix spawn with half energy
- Fix amount of energy set when spawning with spawn tile
- Fix inventory swap objects of the same type (important for entities)
- Add hook calls to onRemove/AddFragment when forcefield is depleted


packages/world/src/systems/ForceFieldSystem.sol
```diff
@@ -140,8 +143,10 @@ contract ForceFieldSystem is System {
     caller.activate();
     caller.requireAdjacentToFragment(fragmentCoord);
 
-    ObjectType objectType = EntityObjectType._get(forceField);
-    require(objectType == ObjectTypes.ForceField, "Invalid object type");
+    {
```
dhvanipa, line 146: why does this cause a stack issue if no new variables were added?

packages/world/src/systems/SpawnSystem.sol
```diff
@@ -142,10 +142,12 @@ contract SpawnSystem is System {
     (EntityId forceField,) = ForceFieldUtils.getForceField(spawnCoord);
     require(!forceField.exists(), "Cannot spawn in force field");
 
+    uint128 spawnEnergy = MAX_PLAYER_ENERGY / 2;
```
dhvanipa, line 145: lool we were removing half but giving full

packages/world/src/systems/TransferSystem.sol
```diff
@@ -28,22 +28,21 @@ contract TransferSystem is System {
...
+    bool callerIsTo = (caller == to);
+
     EntityId target;
 
-    if (from == to) {
-      // Transferring within the same inventory
-      if (caller != from) {
-        target = from;
-      }
+    if (selfTransfer) {
```
dhvanipa, line 37: nice much easier to read with these variable names! nice comments too

dhvanipa (approved):
btw not reviewed the actual tests itself cuz dont think thats needed, just the systems themselves

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -278,9 +278,8 @@ library InventoryUtils {
 
       InventorySlotData memory destSlot = InventorySlot._get(to, slotTo);
 
-      // Handle slot swaps (transferring all to an existing slot with a different type)
-      if (amount == sourceSlot.amount && sourceSlot.objectType != destSlot.objectType && !destSlot.objectType.isNull())
-      {
+      // Handle slot swaps (transferring all to an existing non-empty slot)
+      if (amount == sourceSlot.amount && !destSlot.objectType.isNull()) {
```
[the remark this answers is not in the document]
vdrg: because if lets say you had 2 picks, you wouldn't be able to swap between them if they are the same type, but maybe you want to swap them because one has more mass or something. But now that I think about it, you won't be able to combine stacks of non-tools.. will try to fix in a different way

packages/world/src/systems/ForceFieldSystem.sol
vdrg: ah I think I can remove, it was before when I was doing the program calls inlined, but now they are in a separate function.

packages/world/src/utils/InventoryUtils.sol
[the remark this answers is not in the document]
dhvanipa: > you won't be able to combine stacks of non-tools
ahh oof yea we should keep this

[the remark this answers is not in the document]
dhvanipa: we could check stackable here

PR #256, opened 2025-05-09
Use single EntityPosition table

vdrg:
Might need client changes


dhvanipa:
interesting, the way we generate entity ids for terrain, players, fragments is all via one table so we know they'll be unique

PR #258, opened 2025-05-11
Fix landing on water

vdrg:
- Refactored move lib
- Fixed off by one error when computing path result
- Fixed energy calculation when landing on water
- Don't decrease energy in falls (below threshold)

### DIFF
--- a/packages/world/src/systems/libraries/MoveLib.sol
+++ b/packages/world/src/systems/libraries/MoveLib.sol
@@ -8,8 +8,8 @@ import {
   MAX_PLAYER_GLIDES,
   MAX_PLAYER_JUMPS,
   MOVE_ENERGY_COST,
-  PLAYER_FALL_DAMAGE_THRESHOLD,
-  PLAYER_FALL_ENERGY_COST
+  PLAYER_FALL_ENERGY_COST,
+  PLAYER_SAFE_FALL_DISTANCE
 } from "../../Constants.sol";
 import { EntityId } from "../../EntityId.sol";
 import { ObjectType } from "../../ObjectType.sol";
@@ -30,11 +30,11 @@ library MoveLib {
     uint128 currentEnergy = Energy._getEnergy(player);
 
     uint128 totalCost;
-    Vec3 currentCoord = playerCoord;
+    Vec3 current = playerCoord;
     for (uint256 i = 0; i < newBaseCoords.length; i++) {
-      Vec3 nextBaseCoord = newBaseCoords[i];
-      _requireValidMove(currentCoord, nextBaseCoord);
-      currentCoord = nextBaseCoord;
+      Vec3 next = newBaseCoords[i];
+      _requireValidMove(current, next);
+      current = next;
       totalCost += MOVE_ENERGY_COST;
 
       if (totalCost >= currentEnergy) {
@@ -43,12 +43,14 @@ library MoveLib {
       }
     }
 
-    _setPlayerPosition(playerEntityIds, currentCoord);
+    _setPlayerPosition(playerEntityIds, current);
 
     if (totalCost > 0) {
-      decreasePlayerEnergy(player, currentCoord, totalCost);
-      addEnergyToLocalPool(currentCoord, totalCost);
+      decreasePlayerEnergy(player, current, totalCost);
+      addEnergyToLocalPool(current, totalCost);
     }
+
+    _handleAbove(player, playerCoord);
   }
 
   function move(Vec3 playerCoord, Vec3[] memory newBaseCoords) public {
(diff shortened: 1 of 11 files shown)


packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -19,9 +18,87 @@ import { ObjectTypes } from "../../ObjectType.sol";
...
+      if (totalCost >= currentEnergy) {
+        totalCost = currentEnergy;
+        break;
+      }
+    }
+
+    _setPlayerPosition(playerEntityIds, currentCoord);
+
+    if (totalCost > 0) {
+      decreasePlayerEnergy(player, currentCoord, totalCost);
```
dhvanipa, line 49: Do we have a similar problem of reducing less from the player than total cost, and then adding more to the pool?

dhvanipa: oh wait I guess now `decreasePlayerEnergy` reverts if player has less than totalCost, so not needed

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -19,9 +18,87 @@ import { ObjectTypes } from "../../ObjectType.sol";
...
+
+  function move(Vec3 playerCoord, Vec3[] memory newBaseCoords) public {
+    EntityId[] memory playerEntityIds = _removePlayerPosition(playerCoord);
+    EntityId player = playerEntityIds[0];
+
+    uint128 currentEnergy = Energy._getEnergy(player);
+
+    (Vec3 finalCoord, uint128 totalCost) = _computePathResult(playerCoord, newBaseCoords, currentEnergy);
+
+    if (totalCost > currentEnergy) {
```
dhvanipa, line 62: why is it `>` but `>=` in `moveWithoutGravity`

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -19,9 +18,87 @@ import { ObjectTypes } from "../../ObjectType.sol";
...
+        break;
+      }
+    }
+
+    _setPlayerPosition(playerEntityIds, currentCoord);
+
+    if (totalCost > 0) {
+      decreasePlayerEnergy(player, currentCoord, totalCost);
+      addEnergyToLocalPool(currentCoord, totalCost);
+    }
```
dhvanipa, line 51: why dont we call `_handleAbove` in `moveWithoutGravity`, we're not applying gravity to the current player but there could be a player above them?

dhvanipa:
> Don't decrease energy in falls (below threshold)

Do we need to increase the fall threshold, maybe to 5? Or maybe not because it wasnt being correctly applied before?

packages/world/src/systems/libraries/MoveLib.sol
vdrg: nice catch!

vdrg:
re: fall threshold, I'll rename to safe fall distance and keep it at three, so you will get damage for falls >= 4 (same as minecraft)

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: ahh and theres no loop to break out of here

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -85,165 +145,102 @@ library MoveLib {
...
-  function moveWithoutGravity(EntityId player, Vec3 playerCoord, Vec3[] memory newBaseCoords) public {
-    Vec3[] memory playerCoords = ObjectTypes.Player.getRelativeCoords(playerCoord);
-    EntityId[] memory players = _getPlayers(player, playerCoords);
+      int32 dy = next.y() - current.y();
 
-    // Remove the current player from the grid
-    for (uint256 i = 0; i < playerCoords.length; i++) {
-      ReverseMovablePosition._deleteRecord(playerCoords[i]);
-    }
+      if (dy < 0) {
```
dhvanipa, line 169: hmm without the check on `gravityApplies` dy < 0 doesnt imply a fall right

eg you're moving down a staircase
current: x, y, z
next: x - 1, y - 1, z

vdrg: Also in that case the checks for valid moves would need to allow diagonal movement

dhvanipa: not stairs as in blocks, I mean just stepping down one block

vdrg: Oh shit hadn't realized that, ok will fix and add some more tests

vdrg: hmm well dy < 0 without gravity still might count as a fall, for example at the end of a fall gravity doesn't apply at the last block

dhvanipa (approved):

PR #260, opened 2025-05-12
dont continue mining if tool mass 0

dhvanipa:
- this is so you dont die currently if you're holding a tool

### DIFF
--- a/packages/world/src/systems/MineSystem.sol
+++ b/packages/world/src/systems/MineSystem.sol
@@ -74,7 +74,9 @@ contract MineSystem is System {
       // TODO: factor out the mass reduction logic so it's cheaper to call
       EntityId entityId = _mine(caller, coord, toolSlot, extraData);
       massLeft = Mass._getMass(entityId);
-    } while (massLeft > 0 && Energy._getEnergy(caller) > 0);
+    } while (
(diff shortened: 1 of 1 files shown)


packages/world/src/systems/MineSystem.sol
```diff
@@ -74,7 +74,9 @@ contract MineSystem is System {
       // TODO: factor out the mass reduction logic so it's cheaper to call
       EntityId entityId = _mine(caller, coord, toolSlot, extraData);
       massLeft = Mass._getMass(entityId);
-    } while (massLeft > 0 && Energy._getEnergy(caller) > 0);
+    } while (
+      massLeft > 0 && Energy._getEnergy(caller) > 0 && InventoryUtils.getToolData(caller, toolSlot).massLeft > 0
```
vdrg, line 78: will this work if mining with hand? massLeft will be 0

vdrg: ah I guess it will just mine once

dhvanipa: oh wait true, with hand it should keep mining lol

dhvanipa: actually this is only in the method that passes in a tool slot, we dont pass one in to the hand case

vdrg (approved):

PR #261, opened 2025-05-13
Entity id encoding

vdrg:
- Remove Player, ReversePlayer, ReverseFragmentPosition, ReverseTerrainPosition tables
- Encode player's address in entity id
- Encode fragment and terrain positions in entity ids


packages/world/mud.config.ts
dhvanipa, line 1: Why can we remove the other reverse position tables but not `ReverseMovablePosition`?

vdrg: because for movable entities the position is not stored in the entity id

packages/world/src/EntityId.sol
```diff
@@ -76,6 +77,10 @@ library EntityIdLib {
   }
 
   function getPosition(EntityId self) internal view returns (Vec3) {
+    (EntityType entityType, bytes31 data) = EntityTypeLib.decode(self);
+    if (entityType == EntityTypes.Block || entityType == EntityTypes.Fragment) {
+      return Vec3.wrap(uint96(uint256(bytes32(data) >> 160)));
```
dhvanipa, line 82: why >> 160?

vdrg: just for the aligning to work correctly

packages/world/src/EntityId.sol
```diff
@@ -87,13 +92,36 @@ library EntityIdLib {
     return EntityObjectType._get(self);
   }
 
-  function exists(EntityId self) internal pure returns (bool) {
-    return EntityId.unwrap(self) != bytes32(0);
+  function exists(EntityId self) internal view returns (bool) {
+    return !self.getObjectType().isNull();
```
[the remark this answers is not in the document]
dhvanipa: or do we need 0 for `Incremental`? can `Incremental` be 1?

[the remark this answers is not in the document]
vdrg: no, because knowing if an entity exists or not is independent of the entity id itself, for example if you are referring to the object at coord C, its entity id is encoded but it is not null

packages/world/src/systems/ActivateSystem.sol
```diff
@@ -33,9 +32,7 @@ contract ActivateSystem is System {
   }
 
   function activatePlayer(address playerAddress) public {
-    EntityId player = Player._get(playerAddress);
-    player = player.baseEntityId();
-    require(player.exists(), "Entity does not exist");
+    EntityId player = EntityIdLib.encodePlayer(playerAddress);
```
dhvanipa, line 35: I assume the entity id encoded with the address is always the base entity id

packages/world/src/systems/CraftSystem.sol
```diff
@@ -29,7 +29,6 @@ contract CraftSystem is System {
     require(recipe.inputTypes.length > 0, "Recipe not found");
 
     if (!recipe.stationTypeId.isNull()) {
-      require(station.exists(), "This recipe requires a station");
```
dhvanipa, line 32: Why dont we need to check this? is it because checking the object type is enough?

packages/world/src/systems/ActivateSystem.sol
vdrg: yes, we use an incremental one for the relative coord. Another option is to actually use the same entity id for both, and just map the different coords to it

packages/world/src/systems/CraftSystem.sol
vdrg: yep

packages/world/src/systems/NatureSystem.sol
```diff
@@ -75,15 +74,13 @@ contract NatureSystem is System {
 
     // This is enough to respawn the resource block, as it will be read from the original terrain next time
     EntityObjectType._deleteRecord(entityId);
-    EntityPosition._deleteRecord(entityId);
```
dhvanipa, line 78: Why dont we delete from `EntityPosition` here?

packages/world/src/systems/NatureSystem.sol
```diff
@@ -142,30 +140,29 @@ contract NatureSystem is System {
     uint256 rand = uint256(keccak256(abi.encodePacked(block.timestamp, baseCoord)));
 
     uint32 leafCount;
-    ObjectType leafType = treeData.leafType;
 
     for (uint256 i = 0; i < fixedLeaves.length; ++i) {
       Vec3 rel = fixedLeaves[i];
-      if (rel.y() > int32(trunkHeight)) {
+      if (obstructed && rel.y() > int32(trunkHeight)) {
```
dhvanipa, line 146: was this a bug before? ie not having `obstructed` in the conditional

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -108,7 +81,7 @@ library PlayerUtils {
     if (ReverseMovablePosition._get(coord) != player) {
       return;
     }
-    (EntityId to,) = getOrCreateEntityAt(coord);
+    (EntityId to,) = EntityUtils.getBlockAt(coord);
```
dhvanipa, line 84: Why dont we create the block here like before? ie use `getOrCreateBlockAt`

packages/world/mud.config.ts
dhvanipa: right cuz it keeps changing

packages/world/src/EntityId.sol
dhvanipa: magic math

packages/world/src/EntityId.sol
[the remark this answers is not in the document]
dhvanipa: ahh the entity technically always "exists" lol since its derived from the coord/address. So you have to check another table to see if its set

dhvanipa (approved):
lgtm!

to me the difference is before the black box of getting an entity id used to increment to get the id, and now its encoding in certain cases

everything else has remained the same in this PR: entity ids still store position and object type

packages/world/src/systems/NatureSystem.sol
vdrg: because the entity is always in the same position for terrain

packages/world/src/utils/PlayerUtils.sol
vdrg: because the entity should exist if the player is already there

dhvanipa: does it? I though when we move players, we're not actually creating the entity. I think we're only setting the player position, and reading object type, but not actually creating the entity

vdrg: oh wait you are right, will add a comment

PR #264, opened 2025-05-13
Energy costs and seed balancing

vdrg:


packages/world/src/Constants.sol
```diff
@@ -28,8 +28,10 @@ uint128 constant PLAYER_ENERGY_DRAIN_RATE = 1351521164000;
 
 uint128 constant MACHINE_ENERGY_DRAIN_RATE = 9488203935;
 
-uint128 constant MINE_ENERGY_COST = 8100000000000000;
-uint128 constant HIT_ENERGY_COST = 8100000000000000;
+uint128 constant DEFAULT_MINE_ENERGY_COST = 8100000000000000;
```
dhvanipa, line 31: instead of default should we call this `HAND`?

dhvanipa (approved):

packages/world/src/Constants.sol
dhvanipa: I guess you could be holding a block and mining/hitting, so default makes more sense lol

PR #266, opened 2025-05-14
Seed energy getter

vdrg:
move seed energy to a getter instead of an object, as it makes more sense conceptually (the seed itself doesn't have energy, only the crop/tree does)


dhvanipa (approved):

dhvanipa:
> move seed energy to a getter instead of an object

it's still an object in objects.ts and has an id which is used when building/mining, this is so just so we dont add it the object physics table, correct?

vdrg:
Yeah

PR #271, opened 2025-05-16
add pos hud app

dhvanipa:
- app that shows you your position
- also an example of an app that is not tied to a smart entity


karooolis (approved):

karooolis:
looking good!

PR #275, opened 2025-05-19
Default spawn tile program

vdrg:
should we move the default system external functions to a different system?


dhvanipa (approved):

dhvanipa:
> should we move the default system external functions to a different system?

is this to deal with system too big isuses?

vdrg:
Noo it's so we can call those functions through the world as programs are private

vdrg:
But we can do in a different pr

dhvanipa:
> Noo it's so we can call those functions through the world as programs are private

ahh right, yea that seems like a good idea! then they can show up in the world explorer as well

PR #278, opened 2025-05-19
Enable gas report in ci

vdrg:


packages/world/test/Spawn.t.sol
```diff
@@ -57,7 +57,9 @@ contract SpawnTest is DustTest {
     LocalEnergyPool.set(shardCoord, MAX_PLAYER_ENERGY);
 
     vm.prank(alice);
+    startGasReport("randomSpawn");
```
dhvanipa, line 60: can we also add it to the test that uses a spawn tile?

dhvanipa (approved):

PR #280, opened 2025-05-21
Refactor orientation

vdrg:
Mimics the client in its use of permutations and reflections to provide 48 possible orientations, sometimes called Octahedral group or Cube group: see https://grok.com/share/bGVnYWN5_06152cfb-b8f9-4c2e-944e-9d318219d139


dhvanipa:
also do we have tests for `buildWithOrientation`? If not, we should add at least one test where we build a bed, and then mine it and the right coordinates are cleared

vdrg:
> also do we have tests for `buildWithOrientation`? If not, we should add at least one test where we build a bed, and then mine it and the right coordinates are cleared (for 2 different orientations)

I did add a test for getting relative coords for beds in all supported orientations, but yeah can add for build system too

packages/world/src/ObjectType.sol
```diff
@@ -1155,24 +1155,51 @@ library ObjectTypeLib {
     coords[0] = baseCoord;
 
     for (uint256 i = 0; i < schemaCoords.length; i++) {
-      require(isDirectionSupported(self, direction), "Direction not supported");
-      coords[i + 1] = baseCoord + schemaCoords[i].rotate(direction);
+      require(isOrientationSupported(self, orientation), "Orientation not supported");
```
dhvanipa, line 1158: we should move this outisde the if statement because rn its not getting called for single voxels

vdrg: nice catch, will add a test

PR #286, opened 2025-05-23
Add existence checks to forcefield utils

vdrg:


packages/world/src/utils/ForceFieldUtils.sol
```diff
@@ -104,6 +104,7 @@ library ForceFieldUtils {
    */
   function destroyForceField(EntityId forceField) internal {
```
dhvanipa, line 105: Is this getting used anywhere?

dhvanipa (approved):

packages/world/src/utils/ForceFieldUtils.sol
vdrg: it should 🫠 I'm replacing it with direct calls to the tables, as it should apply to all machines

PR #288, opened 2025-05-23
Optimize inventory

vdrg:


dhvanipa (approved):
I like that now inventory is just 2 tables vs 3, and `InventoryBitmap` is really just there to help in operations!

Also easier to reason about a bitmap specifying which slots are occupied vs nextSlot and occupiedSlots lol

A lot of bit shifting happening in places but I'm glad we have tests so that with an implementation change like this we can still be confident things work

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -131,32 +195,33 @@ library InventoryUtils {
...
+    for (uint256 i = 0; i < bitmap.length && remaining > 0; i++) {
+      uint256 word = bitmap[i];
+      while (word != 0 && remaining > 0) {
+        uint256 bitIndex = LibBit.ffs(word);
+        word &= word - 1; // Clear the bit
+
+        uint16 slot = uint16(i * SLOTS_PER_WORD + bitIndex);
+        InventorySlotData memory data = InventorySlot._get(owner, slot);
+
+        if (data.objectType == objectType && data.amount < stackable) {
```
dhvanipa, line 210: nit: its easier to read if you continue early, like what was there before:
```
if (currentAmount >= stackable) continue;
```

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -166,94 +231,93 @@ library InventoryUtils {
...
-    require(numTypeSlots > 0, "Not enough objects of this type in inventory");
-
     uint16 remaining = amount;
+    uint256[] memory bitmap = InventoryBitmap._getBitmap(owner);
 
-    // Iterate from end to minimize array shifts
-    for (uint256 i = numTypeSlots; i > 0 && remaining > 0; i--) {
-      uint16 slot = InventoryTypeSlots._getItem(owner, objectType, i - 1);
-      uint16 currentAmount = InventorySlot._getAmount(owner, slot);
+    // walk backwards so clearing whole trailing words is cheaper
```
dhvanipa, line 275: why is it backwards here but not in `addObject`

packages/world/src/utils/InventoryUtils.sol
dhvanipa: similar to how you have it in removeObject:

```
if (slotData.objectType != objectType) continue;
```

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -317,145 +380,188 @@ library InventoryUtils {
...
-    for (uint256 i = 0; i < slots.length; i++) {
-      InventorySlotData memory slotData = InventorySlot._get(from, slots[i]);
+      InventorySlotData memory sourceSlot = InventorySlot._get(from, slotFrom);
+      require(!sourceSlot.objectType.isNull(), "Empty slot");
+      fromSlotData[i] = SlotData(sourceSlot.entityId, sourceSlot.objectType, amount);
 
-      if (slotData.entityId.exists()) {
-        // Handle entities (always amount=1)
-        addEntity(to, slotData.entityId);
+      require(from != to, "Cannot transfer to self");
```
dhvanipa, line 399: we can still move stuff around in your own inventory right? do we have a test for that?

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -317,145 +380,188 @@ library InventoryUtils {
...
-    EntityId entityId,
-    ObjectType newObjectType,
-    uint16 amount
-  ) internal {
-    _removeFromTypeSlots(owner, objectType, slot);
-    _addToTypeSlots(owner, newObjectType, slot);
-    InventorySlot._setEntityId(owner, slot, entityId);
-    InventorySlot._setObjectType(owner, slot, newObjectType);
-    InventorySlot._setAmount(owner, slot, amount);
+  function hasObjectType(EntityId owner, ObjectType objectType) internal view returns (bool) {
```
dhvanipa, line 489: Are these utils just here for reference? I dont think we're using them yet:
- hasObjectType
- countObjectsOfType
- getSlotsWithType
- getOccupiedSlotCount
- findObjectType

packages/world/src/utils/InventoryUtils.sol
```diff
@@ -317,145 +380,188 @@ library InventoryUtils {
...
-    ObjectType objectType = InventorySlot._getObjectType(owner, slot);
+  /// @dev Removes empty words from the end of the bitmap.
+  function _pruneTrailingEmptyWords(EntityId owner) private {
+    uint256 length = InventoryBitmap._length(owner);
+    while (length != 0) {
+      uint256 lastWord = InventoryBitmap._getItem(owner, length - 1);
+      if (lastWord != 0) break;
+      InventoryBitmap._popBitmap(owner);
+      unchecked {
+        --length;
```
dhvanipa, line 551: out of curiosity, is this cheaper gas wise because it doesnt do the underflow check (and similarly overflow check in `_ensureWordExists`?

packages/world/src/utils/InventoryUtils.sol
dhvanipa: ahh np, I dont mind them here, up to you

packages/world/src/utils/InventoryUtils.sol
vdrg: yes, we could do it everywhere but it is just ugly, here I felt it was fine

packages/world/src/utils/InventoryUtils.sol
vdrg: because we prune the trailing words if they are empty

PR #290, opened 2025-05-23
[WIP] Ore multipliers

vdrg:


packages/world/src/systems/MineSystem.sol
```diff
@@ -341,12 +345,105 @@ library MineLib {
...
+      objectType = ObjectTypes.Dirt;
+    }
+
+    // Add base type as a drop for all objects
+    result[result.length - 1] = ObjectAmount(objectType, 1);
+
+    return result;
+  }
+
+  function _getRandomDrops(EntityId mined, ObjectType objectType) private view returns (RandomDrop[] memory drops) {
```
dhvanipa, line 380: why was this moved back into the system vs kept in the lib?

dhvanipa (approved):

packages/world/src/systems/MineSystem.sol
vdrg: Because it was only used here

PR #295, opened 2025-05-27
feat: add moderators that can pause the world

vdrg:
- Moderators table
- ModeratorSystem
  - pause/unpause functions by either moderators or admin
  - functions to add/remove moderators by the admin

### DIFF
--- a/packages/world/src/systems/ModeratorSystem.sol
+++ b/packages/world/src/systems/ModeratorSystem.sol
@@ -23,11 +23,11 @@ contract ModeratorSystem is System {
   }
 
   function pause() public onlyModerator {
-    WorldStatus._setInMaintenance(true);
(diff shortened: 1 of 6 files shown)


packages/world/src/systems/ModeratorSystem.sol
```diff
@@ -0,0 +1,46 @@
...
+  modifier onlyModerator() {
+    if (!Moderators._get(_msgSender())) {
+      _requireAdmin();
+    }
+
+    _;
+  }
+
+  function pause() public onlyModerator {
+    WorldStatus._setInMaintenance(true);
```
dhvanipa, line 26: Should we rename `InMaintenance` to something more generic? like paused?

dhvanipa (approved):

PR #302, opened 2025-05-28
feat: fluids

vdrg:
- New EntityFluidLevel table
- Set fluid level when reading a block from terrain
- SpawnsWithFluid category that includes water, magma and underwater plants so we set the initial fluid level to 15
- Replace water checks with fluidLevel > 0
- Make magma not mineable (should we rename back to lava?)


packages/world/src/systems/BedSystem.sol
```diff
@@ -93,38 +81,50 @@ contract BedSystem is System {
...
+  function getPlayerBed(EntityId player) public view returns (EntityId, Vec3) {
+    EntityId bed = PlayerBed._getBedEntityId(player);
+    require(bed._exists(), "Player is not in a bed");
+
+    Vec3 bedCoord = bed._getPosition();
+    return (bed, bedCoord);
+  }
+
+  function transferInventory(EntityId player, EntityId bed) public {
+    InventoryUtils.transferAll(player, bed);
```
dhvanipa, line 128: this confused me for a bit cuz I thought we were transferring to air, and we are

Could we just remove this util in bedlib and call `InventoryUtils.transferAll(player, drop);` directly?

vdrg: bed system is almost over the gas limit :/ but I'll rename the args

dhvanipa: I think `transferAll` is already public though, so it shouldnt change gas no?

packages/world/src/systems/MineSystem.sol
```diff
@@ -171,7 +172,8 @@ contract MineSystem is System {
       return;
     }
 
-    EntityObjectType._set(entityId, ObjectTypes.Air);
+    ObjectType replacementType = EntityFluidLevel._get(entityId) > 0 ? ObjectTypes.Water : ObjectTypes.Air;
```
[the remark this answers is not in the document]
vdrg: yeah

dhvanipa (approved):
🌊 🌋

packages/world/mud.config.ts
```diff
@@ -115,6 +115,14 @@ export default defineWorld({
       },
       key: ["entityId"],
     },
+    // TODO: merge with object type?
```
dhvanipa, line 118: imo shouldnt merge it, better that its separate

vdrg:
@dhvanipa should we remove mass from lava?

dhvanipa:
> @dhvanipa should we remove mass from lava?

yea make it same as water

PR #303, opened 2025-05-29
fix: remove mangrove roots from leaf category

vdrg:
mining these was failing because we don't have a leaf drop chance defined


alvrs:
do we also have to remove them from `ExtraDrops`? https://github.com/dustproject/dust/blob/ee62de9cfbd620b8ef7afb5a6e5d828d31c3035f/packages/world/ts/objects.ts#L953-L956

also someone mentioned this happened for leaves that "pink flowers", i assume they mean flowering azalea leaves - do we have to remove those too, or define a leaf drop chance?

alvrs (approved):

vdrg:
> do we also have to remove them from `ExtraDrops`?
> 
> https://github.com/dustproject/dust/blob/ee62de9cfbd620b8ef7afb5a6e5d828d31c3035f/packages/world/ts/objects.ts#L953-L956
> 
> also someone mentioned this happened for leaves that "pink flowers", i assume they mean flowering azalea leaves - do we have to remove those too, or define a leaf drop chance?

As long as they are not part of the leaf category it should be fine, but I'll remove unused objects from extra drops category

PR #304, opened 2025-05-29
disable script_execution_protection

alvrs:
to allow upgrading foundry


vdrg:
I've been thinking about this, would it be better to instead have a util that returns address(this) bypassing the check? Like a separate contract that just returns msg.sender.

vdrg (approved):

PR #305, opened 2025-05-29
Add azalea leafs to leaf category

vdrg:
(Not added to hasExtraDrops category as there is no sapling)


dhvanipa (approved):
need to codegen object types no?, but lgtm

PR #308, opened 2025-05-30
feat: grow seeds on mine

vdrg:
- Extract growing seeds to NatureLib
- Cleanup mine system
- Grow seeds if mining them or the block below them
- New Landbound object category that determines if the block should be removed if the block below it is being mined


packages/world/src/systems/MineSystem.sol
```diff
@@ -198,7 +198,29 @@ contract MineSystem is System {
...
+    }
+
+    if (aboveType.isLandbound()) {
+      _removeBlock(above, aboveType, aboveCoord);
+      _handleDrop(caller, above, aboveType, aboveCoord);
+    }
+  }
+
+  function _removeGrowable(EntityId entityId, ObjectType objectType, Vec3 coord) internal {
+    EntityObjectType._set(entityId, ObjectTypes.Air);
```
dhvanipa, line 219: why was the check for cannot mine fully grown seed removed here?

packages/world/src/systems/MineSystem.sol
```diff
@@ -198,7 +198,29 @@ contract MineSystem is System {
     }
   }
 
-  function _destroyEntity(EntityId caller, EntityId mined, ObjectType minedType, Vec3 baseCoord) internal {
+  function _handleAbove(EntityId caller, Vec3 coord) internal {
+    // Remove growables on top of this block
+    Vec3 aboveCoord = coord + vec3(0, 1, 0);
+
+    (EntityId above, ObjectType aboveType) = EntityUtils.getOrCreateBlockAt(aboveCoord);
+
```
dhvanipa, line 206: I feel like this should remain a `getBlockAt` instead of `getOrCreateBlockAt`

If you recall, if we keep it `getOrCreateBlockAt`, then on every mine you will be creating an extra entity, so 3 extra table writes

packages/world/src/systems/NatureSystem.sol
```diff
@@ -77,128 +74,10 @@ contract NatureSystem is System {
...
-      (EntityId trunk, ObjectType objectType) = EntityUtils.getOrCreateBlockAt(trunkCoord);
-      if (objectType != ObjectTypes.Air) {
-        return i;
-      }
-
-      EntityObjectType._set(trunk, treeData.logType);
-    }
-
-    return treeData.trunkHeight;
+    // TODO: should we do proximity checks?
```
dhvanipa, line 80: I dont think proximity checks are needed cuz this is just updating the onchain state to reflect what the optimistic value is, cuz theres no tick onchain

packages/world/ts/objects.ts
```diff
@@ -1110,6 +1110,25 @@ export const categories: Record<string, Category> = {
       "MangroveSapling",
     ],
   },
+
+  // Objects that will be mined if they are on top of the block being mined
+  Landbound: {
```
dhvanipa, line 1115: nice, can use this in the client to also optimistically remove these

packages/world/src/systems/MineSystem.sol
vdrg: because now we are always dealing with a non fully grown growable here:
- If the object being mined is growable and is fully grown, we grow it before removing
- If the object above being mined and is fully grown, we also grow it before removing

dhvanipa: gothca

dhvanipa (approved):

PR #310, opened 2025-05-30
chore: prevent building non-waterloggable on water

vdrg:


packages/world/ts/objects.ts
```diff
@@ -1196,6 +1196,103 @@ export const categories: Record<string, Category> = {
...
+      "SprucePlanks",
+      "AcaciaPlanks",
+      "DarkOakPlanks",
+      "MangrovePlanks",
+      "CopperBlock",
+      "IronBlock",
+      "GoldBlock",
+      "DiamondBlock",
+      "NeptuniumBlock",
+      "MangroveSapling",
```
dhvanipa, line 1294: why only one sapling?

PR #311, opened 2025-05-30
feat: dynamic move costs

vdrg:
- Energy drain rate changes depending on final coord (water, lava, normal)
- Move cost over lava is higher (could also add for water)
- Expanded waterloggable category
- TODO: determine constants


packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -226,6 +234,19 @@ library MoveLib {
     }
   }
 
+  function _updatePlayerDrainRate(EntityId player, Vec3 finalCoord) private {
+    uint128 drainRate;
+    if (EntityUtils.getObjectTypeAt(finalCoord - vec3(0, 1, 0)) == ObjectTypes.Lava) {
+      drainRate = PLAYER_LAVA_ENERGY_DRAIN_RATE;
+    } else if (_isFluid(finalCoord + vec3(0, 1, 0))) {
```
dhvanipa, line 241: hmm this only checks fluid level right, so I could end my path on a oak log but still get a swim energy rate?

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -251,4 +272,10 @@ library MoveLib {
   function _isFluid(Vec3 coord) internal view returns (bool) {
     return EntityUtils.getFluidLevelAt(coord) > 0;
   }
+
+  function _getMoveCost(Vec3 coord) internal view returns (uint128) {
+    Vec3 belowCoord = coord - vec3(0, 1, 0);
+    ObjectType belowType = EntityUtils.getObjectTypeAt(belowCoord);
```
dhvanipa, line 278: I think we already do this read during the path check, I wonder if we can save a read by passing it down

packages/world/ts/objects.ts
```diff
@@ -1196,6 +1196,155 @@ export const categories: Record<string, Category> = {
...
+      // "MangroveSapling",
+      // "Furnace",
+      "Workbench",
+      "Powerstone",
+      "ForceField",
+      "Chest",
+      "SpawnTile",
+      // "Bed",
+      "TextSign",
+      // "Torch",
```
dhvanipa, line 1347: torch should be waterloggable

packages/world/src/systems/libraries/MoveLib.sol
vdrg: no because this is the top coord of the player, there can't be anything else there, only passthrough blocks

packages/world/src/systems/libraries/MoveLib.sol
vdrg: yeah, I'm looking into how to optimize this

dhvanipa (approved):

dhvanipa:
I think the client doesnt need any updates because it reads the drain rate from the player to know the optimistic value and that will change post the tx

PR #316, opened 2025-06-03
chore: reachability adjustments

vdrg:
- Use euclidean distance for player range
- Use simple check when mining to see if the mined coord is not covered by other blocks


packages/world/src/Vec3.sol
```diff
@@ -218,6 +236,12 @@ library Vec3Lib {
     return chebyshevDistance(self, other) <= radius;
   }
 
+  function inSphere(Vec3 self, Vec3 other, uint256 radius) internal pure returns (bool) {
+    Vec3 d = other - self;
+    // TODO: should probably use direct subtraction of components to avoid overflow issues
```
dhvanipa, line 241: Seems like we should resolve this todo no? Otherwise we could get overflow?

dhvanipa (approved):

packages/world/src/EntityId.sol
```diff
@@ -83,7 +83,7 @@ library EntityIdLib {
 
   function requireConnected(EntityId self, Vec3 otherCoord) internal view returns (Vec3, Vec3) {
     Vec3 selfCoord = self._getPosition();
-    require(selfCoord.inSurroundingCube(otherCoord, MAX_ENTITY_INFLUENCE_HALF_WIDTH), "Entity is too far");
+    require(selfCoord.inSphere(otherCoord, MAX_ENTITY_INFLUENCE_HALF_WIDTH), "Entity is too far");
```
dhvanipa, line 86: have you tested with the client to see if it causes any issues when trying to mine diagonally?

vdrg: not yet! will do soon

packages/world/src/Vec3.sol
vdrg: no, it is fine for this usecase, it would just happen if the diff between other and self is extremely large, just left it there for later

PR #317, opened 2025-06-04
chore: text sign content

vdrg:
- Move interface to dustkit
- Store text directly in TextSignContent table

### DIFF
--- a/packages/programs/src/programs/TextSignProgram.sol
+++ b/packages/programs/src/programs/TextSignProgram.sol
@@ -5,15 +5,19 @@ import { IBaseWorld } from "@latticexyz/world-consumer/src/experimental/WorldCon
 
 import { EntityId } from "@dust/world/src/EntityId.sol";
 
-import { IDisplay } from "../ClientProgramInterfaces.sol";
-import { ContentURI } from "../codegen/tables/ContentURI.sol";
+import { IDisplay } from "dustkit/IDisplay.sol";
 
+import { TextSignContent } from "../codegen/tables/TextSignContent.sol";
 import { DefaultProgram } from "./DefaultProgram.sol";
 
 contract TextSignProgram is IDisplay, DefaultProgram {
   constructor(IBaseWorld _world) DefaultProgram(_world) { }
 
-  function getContentURI(EntityId, EntityId target, bytes memory) external view returns (string memory) {
-    return ContentURI.get(target);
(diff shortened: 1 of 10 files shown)


packages/programs/src/programs/TextSignProgram.sol
```diff
@@ -5,15 +5,19 @@ import { IBaseWorld } from "@latticexyz/world-consumer/src/experimental/WorldCon
...
   constructor(IBaseWorld _world) DefaultProgram(_world) { }
 
-  function getContentURI(EntityId, EntityId target, bytes memory) external view returns (string memory) {
-    return ContentURI.get(target);
+  function contentURI(EntityId, EntityId target, bytes memory) external view returns (string memory) {
+    string memory content = TextSignContent.get(target);
+    if (bytes(content).length == 0) {
+      return "";
+    }
+    return string.concat("data:text/plain;charset=utf-8,", content);
```
frolic, line 21: we'll need to URI encode `content` for this to work properly

I couldn't find a publicly available lib so this is what ChatGPT wrote for me

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.17;

library URLEncoder {
    function encode(string memory input) internal pure returns (string memory) {
        bytes memory inputBytes = bytes(input);
        bytes memory output = new bytes(inputBytes.length * 3); // Worst case
        uint256 outputLen = 0;

        for (uint256 i = 0; i < inputBytes.length; i++) {
            bytes1 char = inputBytes[i];
            if (
                (char >= "a" && char <= "z") ||
                (char >= "A" && char <= "Z") ||
                (char >= "0" && char <= "9") ||
                char == "-" || char == "_" || char == "." || char == "~"
            ) {
                output[outputLen++] = char;
            } else {
                output[outputLen++] = "%";
                output[outputLen++] = _toHexChar(uint8(char) >> 4);
                output[outputLen++] = _toHexChar(uint8(char) & 0x0f);
            }
        }

        bytes memory encoded = new bytes(outputLen);
        for (uint256 i = 0; i < outputLen; i++) {
            encoded[i] = output[i];
        }

        return string(encoded);
    }

    function _toHexChar(uint8 nibble) private pure returns (bytes1) {
        return nibble < 10 ? bytes1(nibble + 0x30) : bytes1(nibble + 0x41 - 10);
    }
}
```

PR #318, opened 2025-06-04
feat: drop on mine if further than max pickup range

vdrg:
- New constant for max pickup range
- If mine coordinate is in pickup range, it is added to the player's inventory
- If not, it is dropped

### DIFF
--- a/packages/world/src/systems/MineSystem.sol
+++ b/packages/world/src/systems/MineSystem.sol
@@ -44,6 +44,7 @@ import {
   DEFAULT_MINE_ENERGY_COST,
   DEFAULT_ORE_TOOL_MULTIPLIER,
   DEFAULT_WOODEN_TOOL_MULTIPLIER,
+  MAX_PICKUP_RADIUS,
   PLAYER_ENERGY_DRAIN_RATE,
   SAFE_PROGRAM_GAS,
   SPECIALIZED_ORE_TOOL_MULTIPLIER,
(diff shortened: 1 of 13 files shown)


packages/world/src/systems/MineSystem.sol
```diff
@@ -251,16 +252,21 @@ contract MineSystem is System {
...
-    ObjectAmount[] memory result = RandomResourceLib._getMineDrops(mined, minedType, coord);
+    ObjectAmount[] memory result = RandomResourceLib._getMineDrops(mined, minedType, minedCoord);
 
     for (uint256 i = 0; i < result.length; i++) {
       (ObjectType dropType, uint128 amount) = (result[i].objectType, result[i].amount);
 
       if (amount == 0) continue;
 
-      InventoryUtils.addObject(caller, dropType, amount);
+      try InventoryUtils.addObject(caller, dropType, amount) {
```
dhvanipa, line 264: yo what you can try catch in solidity???

dhvanipa (approved):

packages/world/src/systems/MineSystem.sol
dhvanipa: can you use it when calling another contract/

vdrg: Yes, but there are some limitations. It doesn't really catch all issues, for example if there is an issue when decoding the return data, as that happens in the caller contract it will still revert

PR #319, opened 2025-06-06
chore: move improvements

vdrg:
- simplified moveWithoutGravity to just "jump"
- added "move units", which is a way to define a stipend of the moves a player can make per block
- defined unit costs for both walking and swimming, if the sum of these per block surpass the maximum allowed, the path ends and gravity is applied
- unit costs are only applied to non-fall moves (same as energy costs)

Leaving this as a draft until we agree on the values and general approach.


packages/world/src/Constants.sol
```diff
@@ -40,9 +40,14 @@ uint128 constant BUILD_ENERGY_COST = 255500000000000;
 uint128 constant TILL_ENERGY_COST = 255500000000000;
 uint128 constant CRAFT_ENERGY_COST = 255500000000000;
 uint128 constant MOVE_ENERGY_COST = 25550000000000;
-uint128 constant WATER_MOVE_ENERGY_COST = MOVE_ENERGY_COST * 3;
-uint128 constant LAVA_MOVE_ENERGY_COST = MOVE_ENERGY_COST * 10;
-uint128 constant PLAYER_FALL_ENERGY_COST = MAX_PLAYER_ENERGY / 25; // This makes it so, with full energy, you die from a 25 block fall
+uint128 constant WATER_MOVE_ENERGY_COST = MAX_PLAYER_ENERGY / 4000; // 4000 moves in water to die
+uint128 constant LAVA_MOVE_ENERGY_COST = MAX_PLAYER_ENERGY / 10; // 10 moves on lava to die
+uint128 constant PLAYER_FALL_ENERGY_COST = MAX_PLAYER_ENERGY / 25; // This makes it so, with full energy, you die from a 25 + 3 block fall
```
dhvanipa, line 45: could we make this 25 by just setting it to 22?

the other energy cost + drain rates lgtm

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -2,20 +2,11 @@
...
-  MAX_PLAYER_JUMPS,
-  MOVE_ENERGY_COST,
-  PLAYER_ENERGY_DRAIN_RATE,
-  PLAYER_FALL_ENERGY_COST,
-  PLAYER_LAVA_ENERGY_DRAIN_RATE,
-  PLAYER_SAFE_FALL_DISTANCE,
-  PLAYER_SWIM_ENERGY_DRAIN_RATE,
-  WATER_MOVE_ENERGY_COST
-} from "../../Constants.sol";
+import "../../Constants.sol" as Constants;
```
dhvanipa, line 9: woh...

packages/world/test/Move.t.sol
dhvanipa, line 1: we should add some tests

dhvanipa: also related we should add some tests for the new "Coordinate is not reachable"

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -270,15 +266,24 @@ library MoveLib {
...
-  function _getMoveCost(Vec3 coord) internal view returns (uint128) {
+  function _getMoveCost(Vec3 coord) internal view returns (uint128 energyCost, uint128 moveUnitCost) {
     if (EntityUtils.getObjectTypeAt(coord - vec3(0, 1, 0)) == ObjectTypes.Lava) {
-      return LAVA_MOVE_ENERGY_COST;
+      return (Constants.LAVA_MOVE_ENERGY_COST, Constants.WALKING_MOVE_UNIT_COST);
     }
 
-    if (EntityUtils.getObjectTypeAt(coord) == ObjectTypes.Water) {
-      return WATER_MOVE_ENERGY_COST;
+    // TODO: should we instead check below coord?
```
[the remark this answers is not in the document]
vdrg: it is the base, was just thinking that maybe if moving in shallow water (you are in water but with a solid block below) we could just use the regular move cost

packages/world/test/Move.t.sol
vdrg: yeah for the move tests I wanted to know if this approach made sense to others before working on them

PR #320, opened 2025-06-06
Add reachability tests, and make it less strict

vdrg:
- Only check that block has at least one passthrough block as neighbor
- Add tests

### DIFF
--- a/packages/world/src/systems/MineSystem.sol
+++ b/packages/world/src/systems/MineSystem.sol
@@ -104,7 +104,7 @@ contract MineSystem is System {
   function _mine(EntityId caller, Vec3 coord, uint16 toolSlot, bytes calldata extraData) internal returns (EntityId) {
     uint128 callerEnergy = caller.activate().energy;
     caller.requireConnected(coord);
-    MineLib._requireReachable(caller, coord);
+    MineLib._requireReachable(coord);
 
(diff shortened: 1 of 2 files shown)


packages/world/src/systems/MineSystem.sol
```diff
@@ -346,16 +346,14 @@ library MineLib {
...
     for (uint256 i = 0; i < neighbors.length; i++) {
       Vec3 neighbor = neighbors[i];
       ObjectType objectType = EntityUtils.getObjectTypeAt(neighbor);
-      EntityId entity = EntityUtils.getMovableEntityAt(neighbor);
-      bool isEmpty = objectType.isNonSolid() && (entity == caller || !entity._exists());
 
-      // If the neighbor is empty, we consider the coordinate reachable
-      if (isEmpty) return;
+      // If the neighbor is passthrough, we consider the coordinate reachable
+      if (objectType.isPassThrough()) return;
```
dhvanipa, line 356: hmm but players arent pass through right? so then I cant mine a block below someone with this?

dhvanipa: ohh actually the object type would be some passthrough, not player lol

dhvanipa (approved):

PR #321, opened 2025-06-09
Increase move limits

vdrg:
- Reverted decrease in max glides for now
- Made max moves per block 30, and max water moves 90% of that

### DIFF
--- a/packages/world/src/Constants.sol
+++ b/packages/world/src/Constants.sol
@@ -6,7 +6,7 @@ uint32 constant MAX_PICKUP_RADIUS = 6;
 uint32 constant MAX_RESPAWN_HALF_WIDTH = 10;
 
 uint16 constant MAX_PLAYER_JUMPS = 3;
-uint16 constant MAX_PLAYER_GLIDES = 5;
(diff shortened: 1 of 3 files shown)


packages/world/src/Constants.sol
```diff
@@ -6,7 +6,7 @@ uint32 constant MAX_PICKUP_RADIUS = 6;
 uint32 constant MAX_RESPAWN_HALF_WIDTH = 10;
 
 uint16 constant MAX_PLAYER_JUMPS = 3;
-uint16 constant MAX_PLAYER_GLIDES = 5;
+uint16 constant MAX_PLAYER_GLIDES = 10;
```
dhvanipa, line 9: you were seeing reverts from glides?

vdrg: yeah, I think mostly due to path being recomputed + jumping

dhvanipa: I see. I think it'd be good to review the recompute move logic btw. I wrote it wayy back May 2023, but someone else reviewing it will help a lot, and we prob need to update it

PR #323, opened 2025-06-09
pass world address

karooolis:


packages/dustkit/src/rpc/schemas.ts
```diff
@@ -69,6 +69,8 @@ export type AppRpcSchema = RpcSchema.From<{
     method: "dustApp_init";
     params: {
       appConfig: AppConfig;
+      chainId: number;
+      worldAddress: Address;
```
frolic, line 73: we don't use this rpc schema anymore, and instead initialize context here:
https://github.com/dustproject/dust/blob/2745af9adb10346a3bd7f3da1b0ba6441bbf0be6/packages/dustkit/src/bridge/common.ts#L28

packages/dustkit/src/bridge/common.ts
```diff
@@ -26,4 +26,6 @@ export const appContextShape = type({
     },
   ],
   userAddress: hex,
+  worldAddress: hex,
+  chainId: "number",
```
frolic, line 30:
```suggestion
  chainId: "number.integer",
```

PR #324, opened 2025-06-10
Rename hooks, use structs instead of loose args

vdrg:
just for reference, not really necessary


dhvanipa (approved):
nice! I also like the ergonomics of structs better than each individual parameter

I think this will break the current programs though so would have to do in a new world (or migrate existing programs)

PR #325, opened 2025-06-10
feat: add updateProgram function

vdrg:


packages/world/src/systems/ProgramSystem.sol
```diff
@@ -28,24 +28,73 @@ import { IAttachProgramHook, IDetachProgramHook, IProgramValidator } from "../Pr
...
+
+    ProgramId program = target._getProgram();
+    require(program.exists(), "No program attached");
+
+    _detachProgram(caller, target, program, validatorCoord, extraData);
+  }
+
+  function _validateTarget(EntityId target) internal view returns (EntityId) {
     ObjectType targetType = target._getObjectType();
     require(targetType.isSmartEntity(), "Can only attach programs to smart entities");
```
dhvanipa: nit: this is called for detach too, but the revert message says "attach", which could be confusing when debugging

PR #327, opened 2025-06-10
fix: only charge swimming cost if on top of passthrough block with fluid

vdrg:


packages/world/gas-report.json
```diff
@@ -489,7 +489,7 @@
     "file": "test/Tree.t.sol:TreeTest",
     "test": "testTreeGrowthWithSpace",
     "name": "grow oak tree",
-    "gasUsed": 4501392
+    "gasUsed": 4656906
```
vdrg, line 492: no idea why this one changed

PR #332, opened 2025-06-11
rename admin system to debug system

dhvanipa:


packages/world/src/systems/GuardianSystem.sol
```diff
@@ -0,0 +1,46 @@
...
+  function pause() public onlyGuardian {
+    WorldStatus._setIsPaused(true);
+  }
+
+  function unpause() public onlyROOT {
+    WorldStatus._setIsPaused(false);
+  }
+
+  function addGuardian(address moderator) public onlyROOT {
+    require(!Guardians._get(moderator), "Address is already a moderator");
```
dhvanipa, line 34: nit: rename moderator to guardian in revert messages

vdrg: ohh true one sec

PR #337, opened 2025-06-13
Wakeup player if bed is mined

vdrg:


dhvanipa (approved):
lgtm! (lets wait to merge until client is stable though)

PR #338, opened 2025-06-14
Refactor tool utils and reorganize directories

vdrg:


packages/world/src/utils/Math.sol
```diff
@@ -20,8 +20,8 @@ library Math {
     return uint128(FixedPointMathLib.dist(a, b));
   }
 
-  function divUp(uint128 a, uint128 b) internal pure returns (uint128) {
-    return uint128(FixedPointMathLib.divUp(a, b));
+  function divUp(uint256 a, uint256 b) internal pure returns (uint256) {
```
dhvanipa, line 23: why this change? also are we using divUp anywhere?

dhvanipa (approved):
lgtm!

PR #342, opened 2025-06-14
fix: seed tracking

vdrg:
- only track seeds when receiving them as random drops
- add back to circulation when growing


dhvanipa (approved):
we should add the script/query in this PR, so we can run it once we merge this and upgrade the world at the same time

vdrg:
> We should pause the world before running the table migration script and unpause afterwards

yeah of course, we should have a well defined process, ideally utils that do this automatically

PR #343, opened 2025-06-14
Set object mass on grow seed

vdrg:
@dhvanipa I also noticed that dirt has less mass than grass and farmlands, so not sure what the best fix for this is... should we increase dirt's mass and update existing entities? or should we transfer the extra mass to the local pool when doing grass -> farmland -> dirt?

### DIFF
--- a/packages/world/CHANGELOG.md
+++ b/packages/world/CHANGELOG.md
@@ -4,3 +4,4 @@
 
 - Wake up player when mining their bed instead of killing them.
 - Fix: Seeds and saplings in circulation were not being tracked correctly.
+- Fix: Correctly set mass of crops and trees grown from seeds and saplings.


packages/world/src/systems/BuildSystem.sol
```diff
@@ -224,12 +224,16 @@ library BuildLib {
 
     EntityId base = _addBlock(ctx.buildType, ctx.coord);
     EntityOrientation._set(base, ctx.orientation);
+
     Mass._setMass(base, ObjectPhysics._getMass(ctx.buildType));
```
dhvanipa, line 228: Do we need to set this mass here now if we're doing it inside the `EntityUtils.setEntityObjectType`?

dhvanipa: Also do we even need the new call here? We weren't calling `EntityObjectType._set` here before for base

packages/world/src/systems/BuildSystem.sol
```diff
@@ -284,7 +288,5 @@ library BuildLib {
     // if (terrainType == ObjectTypes.Water && !buildType.isWaterloggable()) {
     //   EntityFluidLevel._deleteRecord(terrain);
     // }
-
-    EntityObjectType._set(terrain, buildType);
```
dhvanipa, line 288: Why dont we need this anymore?

dhvanipa:
I think we should do this:
> should we transfer the extra mass to the local pool when doing grass -> farmland -> dirt?

if in the future we want dirt to grow into grass, it would just be getting that energy from the region and X time later, it becomes grass again

packages/world/src/systems/BuildSystem.sol
vdrg: I moved it back to the addBlocks function

packages/world/src/systems/BuildSystem.sol
vdrg: _set was being called in applyTerrainModifications, which I removed. But yes this _setMass is unnecessary as it is already done by setEntityObjectType

PR #350, opened 2025-06-16
Prevent mining forcefields with sleeping players

vdrg:
- We can check if players are sleeping within a forcefield by checking its current drain rate, and prevent mining if that's the case
-  Prevent adding/removing fragments if there are sleeping players in them
- Special case to unblock players that had their forcefield mined while sleeping


packages/world/src/systems/MineSystem.sol
```diff
@@ -294,9 +301,89 @@ contract MineSystem is System {
...
+    * Given: PLAYER_ENERGY_DRAIN_RATE = 1,351,851,852,000
+    *        MACHINE_ENERGY_DRAIN_RATE = 9,488,203,935
+    *        GCD(1351851852000, 9488203935) = 15
+    *
+    * Therefore: N * (1351851852000/15) = K * (9488203935/15)
+    *            N * 90,123,456,800 = K * 632,546,929
+    *
+    * Since 90,123,456,800 and 632,546,929 are coprime (GCD = 1),
+    * N must be a multiple of 632,546,929 for the equation to hold.
+    * This means at least 632,546,929 players must be sleeping in the same forcefield!
```
dhvanipa, line 327: haha I was wondering what N would be for this to be not true, nice proof

packages/world/src/utils/ForceFieldUtils.sol
```diff
@@ -90,12 +91,10 @@ library ForceFieldUtils {
    */
   function removeFragment(EntityId forceField, EntityId fragment) internal {
     require(!fragment._getProgram().exists(), "Can't remove a fragment with a program");
-
-    (EnergyData memory machineData,) = updateMachineEnergy(forceField);
-    Energy._setDrainRate(
-      forceField, machineData.drainRate - MACHINE_ENERGY_DRAIN_RATE - Fragment._getExtraDrainRate(fragment)
-    );
-    Fragment._deleteRecord(fragment);
```
dhvanipa, line 98: Why dont we delete record anymore and instead set forcefield id to null?

dhvanipa (approved):

packages/world/src/utils/ForceFieldUtils.sol
vdrg: because we shouldn't remove the extra drain rate 😞 I'm hoping this hadn't happened yet, as it is kind of a rare edgecase at this point in the game, but we will need to run a migration if it happened.

vdrg: will add an issue to track

packages/world/src/utils/ForceFieldUtils.sol
```diff
@@ -90,12 +91,10 @@ library ForceFieldUtils {
...
+    require(Fragment._getExtraDrainRate(fragment) == 0, "Fragment must not have an extra drain rate");
```
dhvanipa, line 94: so with this you cant expand into fragments with sleeping players?

vdrg: Actually now that we are preventing mining the forcefield, we could still allow this, but might be safer to still prevent it unless we think it is important to support

PR #356, opened 2025-06-18
Add transferAmounts function

vdrg:
Added a transfer function that receives slot amounts instead of slot transfers, so that it is easier for apps to interact with chest programs


dhvanipa (approved):
lgtm

PR #357, opened 2025-06-18
Packed moves

vdrg:
Add moveDirectionsPacked to reduce calldata usage

### DIFF
--- a/packages/world/gas-report.json
+++ b/packages/world/gas-report.json
@@ -15,199 +15,199 @@
     "file": "test/Build.t.sol:BuildTest",
     "test": "testBuildMultiSize",
     "name": "build multi-size",
-    "gasUsed": 229269
+    "gasUsed": 229256
   },
   {
     "file": "test/Build.t.sol:BuildTest",
     "test": "testBuildMultiSizeWithOrientation",
     "name": "build multi-size",
-    "gasUsed": 249455
+    "gasUsed": 249442
   },
   {
     "file": "test/Build.t.sol:BuildTest",
     "test": "testBuildNonTerrain",
     "name": "build non-terrain",
-    "gasUsed": 178992
+    "gasUsed": 179004
   },
   {
     "file": "test/Build.t.sol:BuildTest",
     "test": "testBuildNonWaterloggableBlockOnWaterNonTerrain",
     "name": "build non-waterloggable block on water non-terrain",
-    "gasUsed": 179016
+    "gasUsed": 179028
   },
   {
     "file": "test/Build.t.sol:BuildTest",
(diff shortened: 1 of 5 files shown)


packages/world/gas-report.json
```diff
@@ -273,13 +273,13 @@
     "file": "test/Mine.t.sol:MineTest",
     "test": "testMineResource",
     "name": "mine Ore with hand, entirely mined",
-    "gasUsed": 1035961
+    "gasUsed": 2331846
```
vdrg, line 276: hmm not sure why this one increased..

vdrg: yea just merged it in, but will double check

packages/world/src/systems/MoveSystem.sol
```diff
@@ -14,15 +14,15 @@ import { PlayerUtils } from "../utils/PlayerUtils.sol";
 import { MoveLib } from "./libraries/MoveLib.sol";
 
 contract MoveSystem is System {
-  function move(EntityId caller, Vec3[] memory newCoords) public {
+  function move(EntityId caller, Vec3[] calldata newCoords) public {
```
frolic, line 17: iirc using `calldata` args makes it much harder to call this from another system using the interface? like can't construct a new memory vec3 array and pass it in here?

packages/world/src/systems/MoveSystem.sol
```diff
@@ -36,4 +36,21 @@ contract MoveSystem is System {
 
     notify(caller, MoveNotification({ moveCoords: newCoords }));
   }
+
+  // 5 bits per direction, packed into a single uint256 (max 51 directions)
+  function moveDirectionsPacked(EntityId caller, uint256 packedDirections, uint8 count) public {
```
dhvanipa, line 41: how many directions could we do if we encoded the count into the packed type? maybe as the first thing

dhvanipa (commented):
can we add ts utils so we can call this from the client?

dhvanipa (approved):
also should add changelog

packages/world/src/systems/MoveSystem.sol
vdrg: 49

dhvanipa: maybe worth it cuz then we get one bytes32 instead of 2 right?

packages/world/src/systems/MoveSystem.sol
vdrg: I think it should be easy, but iirc there is a bug in system libs that doesn't allow this so I'll update

PR #359, opened 2025-06-20
[wip] Migrations

### LINKED ISSUE #352: Pending migrations
- Fix mass of existing logs, leaves and crops grown from saplings/seeds. Fix in #347 was already deployed.
- Check if fragments extra drain rate was incorrectly cleared on removal, and if so perform a migration to fix it. The fix to prevent this from happening will be addressed by #350 
- Update seed resource count now that the fix in #346 has been deployed.

vdrg:
- [x] seed tracking
- [ ]  fix growable mass
- [ ] fix drain rates

Fixes #352 

TODO: make sure that the indexer is up to date when performing the migration

### DIFF
--- a/packages/world/script/migrations/1750380758-ƒix-seed-tracking.s.sol
+++ b/packages/world/script/migrations/1750380758-ƒix-seed-tracking.s.sol
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { console } from "forge-std/console.sol";
+
+import { IndexerResult, indexer } from "../utils/Indexer.sol";
+import { Config, config } from "../utils/config.sol";
+
+import { ResourceCount } from "../../src/codegen/tables/ResourceCount.sol";
+import { ObjectType } from "../../src/types/ObjectType.sol";
+
+import { DustScript } from "../DustScript.sol";
+
+contract FixSeedTracking is DustScript {
+  mapping(ObjectType => uint256) private totalCounts;
+
+  function run() public {
+    Config memory cfg = config();
+
+    // Query 1: Get planted seeds/saplings from EntityObjectType
+    console.log("\nCurrent Planted tracking:");
+    IndexerResult memory plantedResult = indexer(cfg).query(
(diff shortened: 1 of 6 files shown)


packages/world/script/migrations/1750380758-ƒix-seed-tracking.s.sol
```diff
@@ -0,0 +1,138 @@
...
+            if (i < seedsAndSaplings.length - 1) {
+                seedTypes = string.concat(seedTypes, ", ");
+            }
+        }
+
+        // Query 1: Get planted seeds/saplings from EntityObjectType
+        console.log("\nCurrent Planted tracking:");
+        IndexerResult memory plantedResult = indexer(cfg).query(
+            string.concat(
+                "SELECT objectType, COUNT(*) as count FROM EntityObjectType WHERE objectType IN (",
```
dhvanipa, line 50: hmm does this mean it's necessarily planted?

packages/world/script/utils/config.sol
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function config() returns (Config memory) {
+  Config memory cfg = _getChainConfig();
+  StoreSwitch.setStoreAddress(address(cfg.world));
+  return cfg;
+}
+
+function _getChainConfig() view returns (Config memory) {
+  if (block.chainid == 690) {
```
dhvanipa, line 21: nice to have: in the future we could read this from `worlds.json` directly since you can read JSON in forge scripts right

packages/world/script/migrations/1750380758-ƒix-seed-tracking.s.sol
dhvanipa: Even if its in your inventory it has the `EntityObjectType`

vdrg: If it is in your inventory it is not an entity

packages/world/script/utils/config.sol
vdrg: Yup

packages/world/script/migrations/1750380758-ƒix-seed-tracking.s.sol
dhvanipa: ahh cuz if you mine it, the entity becomes air

PR #360, opened 2025-06-23
Lazy migration for missing mass

vdrg:
Ended up setting the mass on mine if mass was 0, this should lazily handle missing mass for grown crops/trees.

Also refactored the migration code while attempting to fix mass through a migration.

### DIFF
--- a/packages/world/src/systems/MineSystem.sol
+++ b/packages/world/src/systems/MineSystem.sol
@@ -153,6 +153,13 @@ contract MineSystem is System {
   }
 
   function _prepareBlock(EntityId mined, ObjectType minedType, Vec3 coord) internal returns (ObjectType) {
+    if (Mass._get(mined) == 0) {
+      // If the mass is 0, we assume the block was not correctly setup (e.g. missing mass)
+      // NOTE: This currently targets the issue where grown seeds/saplings were not given mass
+      // TODO: We could potentially stop assigning mass on build and just do it here
+      Mass._setMass(mined, ObjectPhysics._getMass(minedType));
+    }
+
     if (minedType.isMachine()) {
       EnergyData memory machineData = updateMachineEnergy(mined);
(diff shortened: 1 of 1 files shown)


packages/world/src/systems/MineSystem.sol
```diff
@@ -174,6 +174,13 @@ contract MineSystem is System {
       return NatureLib.growSeed(coord, mined, minedType);
     }
 
+    if (Mass._get(mined) == 0) {
```
dhvanipa, line 177: hmm this function has a bunch of early returns, so this wouldnt set the mass for eg for seeds right?

packages/world/src/systems/MineSystem.sol
```diff
@@ -174,6 +174,13 @@ contract MineSystem is System {
...
+      // If the mass is 0, we assume the block was not correctly setup (e.g. missing mass)
+      // NOTE: This currently targets the issue where grown seeds/saplings were not given mass
+      // TODO: We could potentially stop assigning mass on build and just do it here
```
dhvanipa, line 180: Would just move the gas from build to mine right? or are u thinking there'd be some other benefit

packages/world/src/systems/MineSystem.sol
vdrg: yeah, I did it here as for those we know that it was already set.. but maybe it would be more correct to do it at the beginning

packages/world/src/systems/MineSystem.sol
vdrg: yeah no real additional benefit, just less gas when calling getOrCreateBlock, etc

packages/world/src/systems/MineSystem.sol
dhvanipa: yea I think if the idea is that this is a general solution catch-all for all missing mass, then it should be at the beginning imo

PR #362, opened 2025-06-24
Add migration to fix dirt's mass

### LINKED ISSUE #351: Make dirt, grass, farmland and wet farmland mass the same
Currently there is a small mass leak as dirt has less mass than the others. We need a migration to update existing blocks.

vdrg:
also includes some changes due to `pnpm fix`

Fixes #351


dhvanipa (approved):
will need to update in client by bumping `dust` then running `./b script generate_all_biscuits validate`

dhvanipa:
> > > @dhvanipa ooc, what is that needed for?
> > 
> > 
> > Its being used to determine the time to break a block and the durability of a tool, but actually I think we can refactor it to use the mass from dust pkg directly vs copying
> 
> Let me update, should be quick

actually looking into it more, its cuz we store the actual mass as it updates on the item as well, so need the full mass which is read from the item. Will be easier to not change right now lol, just run the script

PR #364, opened 2025-06-24
Make spawn tile passthrough

vdrg:


dhvanipa:
Making them passthrough is more tricky cuz the user would not be able to stand on it, so or simplicity, I made the spawn tiles a full voxel in https://github.com/dustproject/dust-client/pull/978

PR #390, opened 2025-07-07
fix access group ownership

vdrg:
- Allow detaching within custom forcefields (temporary solution)

### DIFF
--- a/packages/programs/src/DefaultProgramSystem.sol
+++ b/packages/programs/src/DefaultProgramSystem.sol
@@ -24,7 +24,7 @@ contract DefaultProgramSystem is System {
 
   function setAccessGroup(EntityId caller, EntityId target, uint256 groupId) external {
     caller.validateCaller();
-    uint256 currentGroupId = EntityAccessGroup.get(target);
+    (uint256 currentGroupId,) = getGroupId(target);
     _requireOwner(currentGroupId, caller);
     EntityAccessGroup.set(target, groupId);
   }
@@ -70,6 +70,6 @@ contract DefaultProgramSystem is System {
   }
 
   function _requireOwner(uint256 groupId, EntityId caller) private view {
-    require(AccessGroupOwner.get(groupId) == caller, "Only the owner can call this function");
+    require(AccessGroupOwner.get(groupId) == caller, "Only the owner of the access group can call this function");
   }
 }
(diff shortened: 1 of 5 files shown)


packages/programs/src/DefaultProgramSystem.sol
```diff
@@ -22,13 +22,23 @@ contract DefaultProgramSystem is System {
...
+  function setAccessGroup(EntityId caller, EntityId target, uint256 groupId) public {
     caller.validateCaller();
-    uint256 currentGroupId = EntityAccessGroup.get(target);
+    (uint256 currentGroupId,) = getGroupId(target);
     _requireOwner(currentGroupId, caller);
     EntityAccessGroup.set(target, groupId);
   }
 
+  /// @dev This function can be called by other entities to get a new access group assigned to themselves
+  function setAccessGroup(EntityId target, address groupOwner) external {
```
frolic, line 33:
```suggestion
  function setAccessGroup(EntityId caller, address groupOwner) external {
```
would this be more accurate?

vdrg: I was just trying to make it a bit clearer that it is setting the group for itself

packages/programs/src/DefaultProgramSystem.sol
```diff
@@ -22,13 +22,23 @@ contract DefaultProgramSystem is System {
     return createAccessGroup(owner);
   }
 
-  function setAccessGroup(EntityId caller, EntityId target, uint256 groupId) external {
+  function setAccessGroup(EntityId caller, EntityId target, uint256 groupId) public {
```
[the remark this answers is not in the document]
vdrg: I believe external still decreases the gas a little bit, though things might have changed recently 🤔 will double check

dhvanipa (approved):
lgtm

packages/programs/src/DefaultProgramSystem.sol
```diff
@@ -22,13 +22,23 @@ contract DefaultProgramSystem is System {
...
     _requireOwner(currentGroupId, caller);
     EntityAccessGroup.set(target, groupId);
   }
 
+  /// @dev This function can be called by other entities to get a new access group assigned to themselves
+  function setAccessGroup(EntityId caller, address groupOwner) external {
+    caller.validateCaller();
+    require(caller.getObjectType().isSmartEntity(), "Target must be a smart entity");
+    (uint256 currentGroupId,) = getGroupId(caller);
+    require(currentGroupId == 0, "Target entity already has an access group");
```
frolic, line 37:
```suggestion
    require(currentGroupId == 0, "Caller already has an access group");
```

PR #394, opened 2025-07-14
set blueprint api

karooolis:


packages/dustkit/src/bridge/schemas.ts
```diff
@@ -61,4 +61,17 @@ export type ClientRpcSchema = RpcSchema.From<
         z: number;
       };
     }
+  | {
+      Request: {
+        method: "setBlueprint";
+        params: {
+          creationId: bigint;
```
dhvanipa, line 68: FYI - this schema will change a bit once I add orientation and remove the old cruft: ie creationId, displayName

(fine for now though)

packages/dustkit/src/bridge/schemas.ts
```diff
@@ -61,4 +61,17 @@ export type ClientRpcSchema = RpcSchema.From<
...
+          objectTypeIds: number[];
+          relativePositions: [number, number, number][];
+          baseWorldCoord: [number, number, number] | undefined;
```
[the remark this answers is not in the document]
frolic: better yet, can the API be

```
blocks: {objectTypeId: number, x: number, y: number, z: number}[]
```

packages/dustkit/src/bridge/schemas.ts
dhvanipa: yea you can make the API match what the contract needs and makes sense

and then stub the values the current client needs, eg creationId

packages/dustkit/src/bridge/schemas.ts
```diff
@@ -61,4 +61,17 @@ export type ClientRpcSchema = RpcSchema.From<
...
     }
+  | {
+      Request: {
+        method: "setBlueprint";
+        params: {
+          objectTypeId: number;
+          x: number;
+          y: number;
+          z: number;
+          rotation: number;
```
dhvanipa, line 72: is this what its called in the contracts? like in the blueprint schema? we should match whatever its called there

karooolis: good point, it's actually `orientation`, updated

PR #408, opened 2025-07-23
Update program hooks

vdrg:
We just need to review the hooks and determine if we want to add any more data


packages/world/src/systems/BuildSystem.sol
```diff
@@ -28,18 +28,17 @@ import { BUILD_ENERGY_COST } from "../Constants.sol";
...
 import { ObjectType, ObjectTypes } from "../types/ObjectType.sol";
 
-import "../ProgramHooks.sol" as Hooks;
 import { ProgramId } from "../types/ProgramId.sol";
 import { Orientation, Vec3, vec3 } from "../types/Vec3.sol";
 
 struct BuildContext {
   EntityId caller;
+  ObjectType slotType;
+  ObjectType buildType;
```
dhvanipa, line 37: why is there both slot and build type? arent they always the same?

packages/world/src/systems/HitMachineSystem.sol
```diff
@@ -39,38 +38,34 @@ contract HitMachineSystem is System {
...
 
     uint128 massReduction = toolData.use(energyLeft, HIT_ACTION_MODIFIER, toolData.toolType.isWhacker());
-
     uint128 machineEnergyReduction = playerEnergyReduction + massReduction;
 
+    Vec3 forceFieldCoord = forceField._getPosition();
     decreaseMachineEnergy(forceField, machineEnergyReduction);
     addEnergyToLocalPool(forceFieldCoord, machineEnergyReduction + playerEnergyReduction);
 
+    caller;
```
dhvanipa, line 62: what is this?

vdrg: To prevent stack too deep

packages/world/src/systems/BuildSystem.sol
vdrg: btw this was here previously, I think I added it when trying to implement the water placement stuff (slot type is water bucket in this case).

packages/world/src/ProgramHooks.sol
```diff
@@ -13,169 +18,265 @@ import { SlotData } from "./utils/InventoryUtils.sol";
...
+    SlotData[] withdrawals;
+  }
+
+  function onTransfer(HookContext calldata ctx, TransferData calldata transfer) external;
 }
 
 interface IHit {
-  function onHit(HitContext calldata ctx) external;
+  struct HitData {
+    uint128 damage;
```
dhvanipa, line 59: for both hits and mines we could include which tool was used

dhvanipa (approved):

packages/world/src/systems/BuildSystem.sol
dhvanipa: maybe we pass in `SlotData` then

vdrg: wdym? to the hook?

packages/world/src/systems/MineSystem.sol
```diff
@@ -164,19 +147,52 @@ contract MineSystem is System {
...
+    Mass._deleteRecord(mined);
+
+    Vec3 baseCoord = mined._getPosition();
+
+    // Get all coords: index 0 is base, rest are relatives
+    Vec3[] memory coords = minedType.getRelativeCoords(baseCoord, EntityOrientation._get(mined));
+
+    for (uint256 i = 0; i < coords.length; i++) {
+      Vec3 coord = coords[i];
+      _handleAbove(caller, coord);
```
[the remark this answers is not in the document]
dhvanipa: gotcha yea that could happen

alvrs (approved):
lgtm

PR #409, opened 2025-07-23
add terrain lib equivalent ts utils

dhvanipa:
Equivalent util to https://github.com/dustproject/dust/blob/main/packages/world/src/systems/libraries/TerrainLib.sol#L38

### DIFF
--- a/packages/world/ts/terrain.ts
+++ b/packages/world/ts/terrain.ts
@@ -82,8 +82,14 @@ export async function getTerrainBlockType(
     chunkCoord,
   );
 
-  const index = getBlockIndex([x, y, z]);
-  return readBytes1(bytecode, index);
+  return readTerrainBlockType(bytecode, [x, y, z]);
+}
+
+export function readTerrainBlockType(
+  chunkBytecode: string,
+  [x, y, z]: Vec3,
+): number {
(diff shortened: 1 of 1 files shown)


packages/world/ts/terrain.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const DATA_OFFSET = 1;
+
+const EXPECTED_VERSION = 0x00;
+
+const VERSION_PADDING = 1;
+const BIOME_PADDING = 1;
+const SURFACE_PADDING = 1;
+
+function _getChunkSalt(coord: Vec3) {
```
vdrg, line 14: remove _ prefix

vdrg (approved):
just a few nits, lgtm otherwise!

packages/world/ts/chunk.ts
```diff
@@ -0,0 +1,15 @@
+import type { ReadonlyVec3, Vec3 } from "./vec3";
+
+export const CHUNK_SIZE = 16;
+
+export function voxelToChunkPos([x, y, z]: ReadonlyVec3): Vec3 {
```
vdrg, line 5: should it also return readonly?

vdrg (approved):

PR #413, opened 2025-07-26
Add bytecode caching to terrain functions

dhvanipa:
Cache chunk bytecode in getTerrainBlockType and getBiome to avoid redundant network calls when accessing the same chunk multiple times.

🤖 Generated with [Claude Code](https://claude.ai/code)


frolic:
> what is this for? shouldn't the user of the utils be the ones implementing the cache?

I suggested this because it simplifies what a user needs to do and reduces load on our RPC for cases where a user doesn't do what they should be doing. Since the data is static/immutable, it should be safe to just cache it for em. (We already do similar caching in a variety of places in MUD and Viem does similar things)

PR #415, opened 2025-07-29
PVP

### LINKED ISSUE #378: enable PvP
* allow hitting other players if within a small range
* only allow hitting with bare hands for now while we don't have armor

### LINKED ISSUE #383: action rate limit
we already rate limit movements, but should prob also rate limit mining, etc.

@vdrg mentioned some generic way to bucket actions so we can have diff actions have diff rate limits

vdrg:
Reachability check through raycasting will be done in a different PR

Closes #378
Closes #383


packages/world/mud.config.ts
```diff
@@ -8,7 +8,7 @@ export default defineWorld({
     generateSystemLibraries: true,
   },
   enums: {
-    Action: [
+    NotificationType: [
```
dhvanipa, line 11: Does renaming this require any migration? or no cuz the enum is just a uint8?

also the field is still action on the table so maybe we should keep it this?

packages/world/mud.config.ts
```diff
@@ -227,13 +229,14 @@ export default defineWorld({
       },
       key: ["x", "y", "z"],
     },
-    MoveUnits: {
```
dhvanipa, line 230: since the table isnt deleted on upgrade, we should keep it but just move it below and add a deprecated comment

packages/world/src/Constants.sol
```diff
@@ -94,3 +87,18 @@ uint256 constant MAX_NEPTUNIUM = 18_696;
...
+// Rate limit constants
+uint128 constant MAX_RATE_LIMIT_UNITS_PER_BLOCK = 1e18;
+uint128 constant MAX_RATE_LIMIT_UNITS_PER_SECOND = MAX_RATE_LIMIT_UNITS_PER_BLOCK / BLOCK_TIME;
+
+// Movement rate limits
+uint128 constant WALK_UNIT_COST = MAX_RATE_LIMIT_UNITS_PER_SECOND / 15; // 15 blocks per second
+uint128 constant SWIM_UNIT_COST = MAX_RATE_LIMIT_UNITS_PER_SECOND * 10 / 135; // 13.5 blocks per second (90% of walking speed)
+
+// Combat rate limits
+uint128 constant HIT_UNIT_COST = MAX_RATE_LIMIT_UNITS_PER_BLOCK; // 1 hit per block
```
dhvanipa, line 100: hmm how will 1 hit per block with preconfs?

packages/world/src/systems/BuildSystem.sol
```diff
@@ -151,56 +157,36 @@ contract BuildSystem is System {
...
-      return (program, fragment);
-    }
-
-    return (forceField._getProgram(), forceField);
-  }
 }
 
 library BuildLib {
   function _executeBuild(BuildContext memory ctx) public returns (EntityId, Vec3[] memory) {
-    (ctx.callerEnergy,) = transferEnergyToPool(ctx.caller, Math.min(ctx.callerEnergy, BUILD_ENERGY_COST));
```
dhvanipa, line 202: hmm didnt we need to set this on the build ctx?

packages/world/src/systems/HitPlayerSystem.sol
```diff
@@ -0,0 +1,94 @@
...
+  ) internal {
+    // Check if target is within a force field and call hooks
+    (ProgramId program, EntityId hookTarget, EnergyData memory energyData) = ForceFieldUtils.getHookTarget(targetCoord);
+
+    if (!program.exists()) {
+      return;
+    }
+
+    program.hook({ caller: caller, target: hookTarget, revertOnFailure: energyData.energy > 0, extraData: extraData })
+      .onHit(tool, totalDamage);
```
dhvanipa, line 92: hmm we're using this for hitting the machine right? How would the program know this is a hit on the player?

packages/world/mud.config.ts
vdrg: it shouldn't require a migration, we can just rename in the table. I don't like action, like Death is not an action

packages/world/src/systems/HitPlayerSystem.sol
vdrg: yes I think we will need to add a new field or have a different hook like onHitPlayer

packages/world/src/Constants.sol
vdrg: it should just revert, we should probably handle the limit in the client

packages/world/src/systems/BuildSystem.sol
vdrg: no, we only need it here, so trying to not mutate the ctx if possible

packages/world/src/utils/RateLimitUtils.sol
```diff
@@ -0,0 +1,54 @@
...
+  // Internal helper
+  function _updateRateLimit(EntityId entity, RateLimitType limitType, uint128 unitCost) private {
+    if (unitCost == 0) {
+      return; // No cost, no rate limit update needed
+    }
+
+    uint128 currentUnits = RateLimitUnits._get(entity, block.number, limitType);
+    uint128 newUnits = currentUnits + unitCost;
+
+    require(newUnits <= MAX_RATE_LIMIT_UNITS_PER_BLOCK, "Rate limit exceeded");
```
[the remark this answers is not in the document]
vdrg: not really, it is just a way to set some value, the important thing will by the unit costs

packages/world/src/Constants.sol
```diff
@@ -41,23 +42,15 @@ uint128 constant PLAYER_LAVA_ENERGY_DRAIN_RATE = MAX_PLAYER_ENERGY / 10 seconds;
 
 uint128 constant MACHINE_ENERGY_DRAIN_RATE = 9488203935;
 
-uint128 constant DEFAULT_MINE_ENERGY_COST = 8100000000000000;
-uint128 constant TOOL_MINE_ENERGY_COST = 255500000000000;
-uint128 constant DEFAULT_HIT_ENERGY_COST = 8100000000000000;
-uint128 constant TOOL_HIT_ENERGY_COST = 255500000000000;
+uint128 constant ACTION_ENERGY_COST = 255500000000000;
```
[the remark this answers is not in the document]
vdrg: I agree, would love your thoughts on this, but basically I refactored all actions that use tools to use the same constant for energy, that way we don't need to pass it around when doing `toolData.use()`. We could just rename this to something more appropriate or rethink the model for this.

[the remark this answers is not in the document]
vdrg: what about 

```
uint128 constant TOOL_ACTION_ENERGY_COST = 255500000000000;
uint128 constant BARE_HANDS_ACTION_ENERGY_COST = 8100000000000000;
```

packages/world/mud.config.ts
vdrg: will add back

packages/world/mud.config.ts
dhvanipa: we can also deprecate this table for a new one where the field isnt called action, up to you though

packages/world/src/systems/HitPlayerSystem.sol
dhvanipa: nice so its now a field on onHit! this will require a migration of existing force field programs right?

packages/world/src/Constants.sol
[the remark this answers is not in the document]
dhvanipa: > what about
> 
> ```
> uint128 constant TOOL_ACTION_ENERGY_COST = 255500000000000;
> uint128 constant BARE_HANDS_ACTION_ENERGY_COST = 8100000000000000;
> ```

I like this!

packages/world/src/systems/HitPlayerSystem.sol
```diff
@@ -0,0 +1,91 @@
...
+
+  function hitPlayer(EntityId caller, EntityId target, bytes calldata extraData) public {
+    _hitPlayer(caller, target, type(uint16).max, extraData);
+  }
+
+  function _hitPlayer(EntityId caller, EntityId target, uint16 toolSlot, bytes calldata extraData) internal {
+    // Update and check caller's energy
+    caller.activate();
+
+    (, Vec3 targetCoord) = caller.requireInRange(target, MAX_HIT_RADIUS);
```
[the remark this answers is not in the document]
dhvanipa: https://github.com/dustproject/dust/blob/vdrg/pvp/packages/world/src/types/EntityId.sol#L83

It doesnt return targetCoord, but coord right now

packages/world/src/utils/PlayerUtils.sol
```diff
@@ -77,10 +77,8 @@ library PlayerUtils {
 
   /// @dev Kills the player, it assumes the player is not sleeping
   // If the player was already killed, it will return early
-  function killPlayer(EntityId player, Vec3 coord) internal {
-    if (ReverseMovablePosition._get(coord) != player) {
```
dhvanipa, line 81: dont we still need to check if the player isnt already dead?

packages/world/src/Constants.sol
```diff
@@ -94,3 +87,21 @@ uint256 constant MAX_NEPTUNIUM = 18_696;
...
+// Rate limit constants
+uint128 constant MAX_RATE_LIMIT_UNITS_PER_BLOCK = 1e18;
+uint128 constant MAX_RATE_LIMIT_UNITS_PER_SECOND = MAX_RATE_LIMIT_UNITS_PER_BLOCK / BLOCK_TIME;
+
+// Movement rate limits
+uint128 constant WALK_UNIT_COST = MAX_RATE_LIMIT_UNITS_PER_SECOND / 15; // 15 blocks per second
+uint128 constant SWIM_UNIT_COST = MAX_RATE_LIMIT_UNITS_PER_SECOND * 10 / 135; // 13.5 blocks per second (90% of walking speed)
+
+// Combat rate limits (separate buckets)
+// Keep legacy HIT_UNIT_COST for compatibility; prefer HIT_PLAYER_UNIT_COST and HIT_MACHINE_UNIT_COST
```
dhvanipa, line 100: Why do we need to keep `HIT_UNIT_COST`? Its not used anywhere no?

PR #416, opened 2025-07-31
Body progression

vdrg:
- Added simple tracking of different activities
- Wooden tools now have to be crafted on a workbench
- Added migration for recipes
- Apply up to 30% energy discount depending on progress

Details:
https://docs.google.com/spreadsheets/d/1eue3MELkw5AL5tqFE29uSEjUBmTd1P6tl3Ii1XHBhLI/edit?usp=sharing


docs/pages/apps/dustkit.md
dhvanipa, line 1: hmm there's a bunch of diff here that seems unrelated to this PR

vdrg: fixed! though I'm working now on the decay stuff

packages/world/script/migrations/3-migrate-recipes/migration.s.sol
```diff
@@ -0,0 +1,92 @@
...
+        string.concat(newRecipeIdsStr, "decode('", vm.replace(vm.toString(newRecipeIds[i]), "0x", ""), "', 'hex')");
+      if (i < newRecipeIds.length - 1) {
+        newRecipeIdsStr = string.concat(newRecipeIdsStr, ", ");
+      }
+    }
+
+    // Query recipes that exist in the table but not in new recipes (to be deleted)
+    console.log("\nQuerying recipes to delete...");
+    string memory deleteQuery =
+      string.concat("SELECT recipeId FROM Recipes WHERE recipeId NOT IN (", newRecipeIdsStr, ") ORDER BY recipeId;");
```
dhvanipa, line 43: is this supported in dozer?

packages/world/src/systems/BuildSystem.sol
```diff
@@ -147,52 +168,53 @@ contract BuildSystem is System {
       SeedGrowth._setFullyGrownAt(base, uint128(block.timestamp) + ctx.buildType.getTimeToGrow());
     } else if (ctx.buildType.hasExtraDrops()) {
       DisabledExtraDrops._set(base, true);
+    } else if (ctx.buildType == ObjectTypes.ForceField) {
+      (EntityId existing,) = ForceFieldUtils.getForceField(ctx.coord);
```
dhvanipa, line 172: Doesn't this check need to run in all coords not just base, since the force field tower could take up more than 1 position? or we're assuming base is where its at, similar to beds?

packages/world/src/systems/BuildSystem.sol
```diff
@@ -147,52 +168,53 @@ contract BuildSystem is System {
...
-    uint128 energyLeft = transferEnergyToPool(ctx.caller, Math.min(ctx.callerEnergy, BUILD_ENERGY_COST));
-    if (energyLeft == 0) {
+    if (transferEnergyToPool(ctx.caller, ctx.energyCost) == 0) {
       return (EntityId.wrap(0), new Vec3[](0));
     }
 
     _updateInventory(ctx);
 
-    return _addBlocks(ctx);
+    (EntityId base, Vec3[] memory coords) = _addBlocks(ctx);
```
dhvanipa, line 215: do we need to bring these variables out? Or can we just return like before

packages/world/src/systems/BuildSystem.sol
```diff
@@ -147,52 +168,53 @@ contract BuildSystem is System {
...
+      extraData: ctx.extraData
+    });
+  }
 }
 
 library BuildLib {
   function _executeBuild(BuildContext memory ctx) public returns (EntityId, Vec3[] memory) {
-    uint128 energyLeft = transferEnergyToPool(ctx.caller, Math.min(ctx.callerEnergy, BUILD_ENERGY_COST));
-    if (energyLeft == 0) {
+    if (transferEnergyToPool(ctx.caller, ctx.energyCost) == 0) {
```
dhvanipa, line 209: Don't we still need to do Math.min? what if player has less energy than the energyCost?

packages/world/src/systems/CraftSystem.sol
```diff
@@ -32,12 +36,18 @@ contract CraftSystem is System {
...
+    uint128 callerEnergy = transferEnergyToPool(caller, effectiveCost);
     require(callerEnergy > 0, "Not enough energy to craft");
 
     CraftLib._consumeRecipeInputs(caller, recipe, inputs);
     CraftLib._createRecipeOutputs(caller, recipe);
-    CraftLib._handleEnergyConservation(caller, recipe);
+    uint128 massEnergy = CraftLib._handleEnergyConservation(caller, recipe);
+
+    // Track crafting activity
+    // TODO: this is the output's mass+energy, should we track input instead?
```
dhvanipa, line 49: yea can we combine input and output?

packages/world/src/systems/HitPlayerSystem.sol
```diff
@@ -66,26 +67,27 @@ contract HitPlayerSystem is System {
...
     Vec3 targetCoord,
     EntityId tool,
     uint128 totalDamage,
     bytes calldata extraData
-  ) internal {
+  ) public {
     // Check if target is within a force field and call hooks
     (ProgramId program, EntityId hookTarget, EnergyData memory energyData) = ForceFieldUtils.getHookTarget(targetCoord);
 
-    if (!program.exists()) {
```
dhvanipa, line 85: Similar question as before, why dont we need this now?

packages/world/src/utils/OreLib.sol
```diff
@@ -47,8 +37,12 @@ library OreLib {
   }
 
   function burnOre(ObjectType self, uint256 amount) internal {
+    uint256 resourceCount = ResourceCount._get(self);
+
+    require(resourceCount >= amount, "Not enough resources to burn");
```
dhvanipa, line 42: ahh nice, this is a better error message than the arithmetic solidity error lol

packages/world/test/InventoryUtils.t.sol
```diff
@@ -304,7 +304,7 @@ contract InventoryUtilsTest is DustTest {
   // Fuzz test for transferring objects between slots
   function testFuzzTransferBetweenSlots(uint16 amount1, uint16 amount2, uint16 transferAmount) public {
     // Bound inputs to reasonable values
-    amount1 = uint16(bound(amount1, 1, 99)); // Max stack is 99
+    amount1 = uint16(bound(amount1, 1, 98)); // Max stack is 99
```
dhvanipa, line 307: should this not be 99?

packages/world/ts/scripts/genGetRecipesSol.ts
dhvanipa, line 1: this replaces `genInitRecipesSol` so we should delete that

packages/world/src/systems/MineSystem.sol
```diff
@@ -79,78 +76,88 @@ contract MineSystem is System {
...
-  }
 
-  function _mine(EntityId caller, Vec3 coord, uint16 toolSlot, bytes calldata extraData) internal returns (EntityId) {
-    uint128 callerEnergy = caller.activate().energy;
+    // Prepare context once (resolves object type, handles ore collapse/growth/machine checks)
+    MineContext memory ctx = _mineContext({ caller: caller, coord: coord, extraData: extraData, toolSlot: toolSlot });
 
-    caller.requireConnected(coord);
-    MineLib._requireReachable(coord);
+    // Tight loop with minimal per-iteration allocations
```
dhvanipa, line 113: nice this prob made mineUntilDestroyed a lot cheaper

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -119,73 +154,92 @@ library MoveLib {
...
+
+    return (current, stepCtx, _computeFallDamage(ctx, stepCtx, currentFallHeight));
+  }
+
+  function _clampEnergyCosts(MoveContext memory ctx, uint128 currentCost, uint128 moveCost, uint128 fallCost)
+    internal
+    pure
+    returns (uint128, uint128)
+  {
+    if (ctx.initialEnergy <= currentCost) {
```
dhvanipa, line 180: does this need to be `<` not `<=` since if its equal we still want to bring it to 0?

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -79,17 +115,16 @@ library MoveLib {
     EntityId[] memory playerEntityIds = _removePlayerPosition(playerCoord);
     EntityId player = playerEntityIds[0];
 
-    (Vec3 finalCoord, uint128 totalCost) = _computeGravityResult(playerCoord, 0);
-
-    _setPlayerPosition(playerEntityIds, finalCoord);
+    MoveContext memory ctx = _moveContext(player);
 
-    uint128 currentEnergy = updatePlayerEnergy(player).energy;
```
dhvanipa, line 86: In move and jumpBuild we can just call getEnergy because we already updated the energy in the movesystem, but when gravity runs the player that's falling might not be the same player as the caller, and will not have their energy updated

Also hmm curious why any of the gravity tests didnt catch this

packages/world/src/utils/PlayerSkillUtils.sol
```diff
@@ -0,0 +1,164 @@
...
+  {
+    ActivityType activityType;
+    if (minedType.isCrop()) {
+      activityType = ActivityType.MineCropMass;
+    } else if (toolType.isAxe()) {
+      activityType = ActivityType.MineAxeMass;
+    } else if (toolType.isPick()) {
+      activityType = ActivityType.MinePickMass;
+    } else {
+      // TODO: use wad
```
dhvanipa, line 50: can we just import Wad from solady?

packages/world/src/utils/ToolUtils.sol
```diff
@@ -28,85 +29,109 @@ struct ToolData {
...
-      // Transfer the player's energy to the local pool
-      uint128 remainingEnergy = decreasePlayerEnergy(toolData.owner, energyCost);
-      addEnergyToLocalPool(callerCoord, energyCost);
+      // Use player specific util to decrease bytecode size
+      decreasePlayerEnergy(toolData.owner, energyCost);
+      addEnergyToLocalPool(toolData.owner._getPosition(), energyCost);
 
-      // If player died, return early
-      if (remainingEnergy == 0) {
+      if (energyCost == callerEnergy) {
```
dhvanipa, line 92: ahh energyCost is already capped to callerEnergy in `getActionEnergyCost` nice

dhvanipa (approved):
besides the comments, lgtm! Lots of big refactors that will make it easier to add new features in the future, nice!

packages/world/src/systems/BuildSystem.sol
vdrg: yes I think we should always keep it at 1, because even if it used more voxels we should only setup a single fragment as the "main" fragment and expand from there. I think we should keep forcefields as a single block forever, otherwise it becomes trickier if the forcefield is in multiple fragments.

packages/world/src/systems/BuildSystem.sol
```diff
@@ -147,52 +168,53 @@ contract BuildSystem is System {
...
         objectType: ctx.buildType,
         orientation: ctx.orientation
       });
     }
   }
+
+  /// @dev Returns the pre-built hook to avoid stack too deep issues
+  function _hook(BuildContext memory ctx, Vec3 coord) internal returns (Hook memory) {
+    (ProgramId program, EntityId target, EnergyData memory energyData) = ForceFieldUtils.getHookTarget(coord);
+    return program.hook({
```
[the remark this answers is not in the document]
vdrg: the hook calls check it internally

packages/world/src/systems/BuildSystem.sol
vdrg: we do that when building the context

packages/world/src/systems/CraftSystem.sol
vdrg: hmm but this would be like 2*input if energy is conserved no? so I think it is either input or output but not both. The main difference is for cases like furnaces and dyes, where input >> output

packages/world/src/systems/HitPlayerSystem.sol
vdrg: when calling the hook it is checked internally

packages/world/src/utils/OreLib.sol
vdrg: yeah, got tired of debugging the same issue over and over when tests broke lol

packages/world/src/utils/PlayerProgressUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+    return uint128(Math.max(current, _floor(accumulated)));
+  }
+
+  // Mining tracking - with tool specialization and crop detection
+  function trackMine(EntityId player, uint128 massReduced, ObjectType toolType, ObjectType minedType) internal {
+    ActivityType activityType;
+
+    // Check if we're mining a crop
+    if (minedType.isCrop()) {
+      // TODO: should we still track axe/pick mining independently?
```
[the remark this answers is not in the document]
vdrg: in this case, how should the discount be applied? combine both crop+tool discount? or to keep it even simpler we could just remove crop mining and figure it out later

packages/world/src/utils/PlayerProgressUtils.sol
```diff
@@ -0,0 +1,165 @@
...
+    _trackProgress(player, ActivityType.MoveEnergy, moveEnergy);
+  }
+
+  function trackFallEnergy(EntityId player, uint128 fallEnergy) internal {
+    _trackProgress(player, ActivityType.MoveFallEnergy, fallEnergy);
+  }
+
+  // Building tracking
+  function trackBuild(EntityId player, uint128 massBuilt) internal {
+    // TODO: should we include build energy? if so, it will be dependent on the energy discount
```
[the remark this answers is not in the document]
vdrg: just for reference: one interesting thing about including build energy, is that as the energy goes down as the progress goes up, the progress curve changes a bit. But I agree, better to keep it simple for now

packages/world/src/utils/PlayerSkillUtils.sol
vdrg: yup I actually wrapped most of solady stuff in our Math lib now, so we don't have to mix both. Also adding WAD to our constants

packages/world/ts/scripts/genGetRecipesSol.ts
vdrg: oh true I thought I had removed it

packages/world/src/systems/libraries/MoveLib.sol
vdrg: yes I think the tests don't catch it because in most tests the block timestamp is the same throughout the test. I need to figure out how to fit the updatePlayerEnergy within the contract size limit 🫠

packages/world/src/systems/libraries/MoveLib.sol
```diff
@@ -119,73 +154,92 @@ library MoveLib {
...
-          (uint128 moveCost, bool isSwimming) = _getMoveCost(next);
-          cost += moveCost;
-
-          // Track movement type for batch update
-          if (isSwimming) {
-            swimSteps++;
-          } else {
-            walkSteps++;
-          }
+          uint128 fallDamage = _computeFallDamage(ctx, stepCtx, falls);
```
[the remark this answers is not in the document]
vdrg: hmm that's a good point. Will think more about this. Quick question, do you think jumps should be allowed while falling?

packages/world/src/systems/CraftSystem.sol
dhvanipa: ahh right yea its conserved so dont need it

packages/world/src/utils/PlayerProgressUtils.sol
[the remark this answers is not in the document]
dhvanipa: yea actually remove crop mining and figure it out later

packages/world/ts/recipes.ts
```diff
@@ -123,6 +123,13 @@ export const recipes: Recipe[] = [
     ],
     outputs: [["Powerstone", 1]],
   },
+  {
+    inputs: [
+      ["IronBar", 3],
```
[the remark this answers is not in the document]
dhvanipa: we could keep it a stone anvil, but allow crafting copper tools at a workbench, that'd work too cuz u can get iron from copper

packages/world/src/systems/libraries/MoveLib.sol
vdrg: did it by moving the responsibility of calling updatePlayerEnergy to the caller system

packages/world/ts/recipes.ts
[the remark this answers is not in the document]
vdrg: hmm but it would be kind of weird to have the special case for copper, might be confusing?

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: ahh I see, makes sense

packages/world/src/systems/libraries/MoveLib.sol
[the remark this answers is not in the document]
dhvanipa: hmm why do we still need the check for `currentHasGravity`?

` uint128 fallDamage = currentHasGravity && !nextHasGravity ? _computeFallDamage(ctx, stepCtx, falls) : 0;`

I'm just comparing it to the old condition, and we didnt have it there: https://github.com/dustproject/dust/blob/main/packages/world/src/systems/libraries/MoveLib.sol#L209

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: hmm, but where is it still called for the gravity case?

dhvanipa: Like this only fixes the player you're mining, but `runGravity` is recursively called:

```
updatePlayerEnergy(above);
MoveLib.runGravity(aboveCoord);
```

dhvanipa: why wont it fit in `runGravity`? It's already a public function no?

packages/world/src/systems/libraries/MoveLib.sol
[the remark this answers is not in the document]
vdrg: before it was checking for all iterations, but we only care about the landing step when applying damage

packages/world/src/systems/libraries/MoveLib.sol
vdrg: ahh it seems that with the latest refactor, MoveLib decreased in size just enough to fit that, so will add it to the moveContext function

packages/world/ts/recipes.ts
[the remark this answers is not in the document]
dhvanipa: yea it would be confusing, actually thinking lets just remove anvil from this PR, can do in a separate one. Also makes it so we dont need any client changes then for body progression

packages/world/src/systems/libraries/MoveLib.sol
dhvanipa: nicee! let's remove the `updateEnergy` from `Minelib._removeBlock` now then

PR #420, opened 2025-08-07
add map tile requests

dhvanipa:


dhvanipa:
not needed, found a better way to do this

PR #425, opened 2025-08-14
gen ore lib

dhvanipa:


packages/world/ts/terrain.ts
```diff
@@ -156,3 +153,24 @@ export function readTerrainBlockType(
 ): number {
   return readBytes1(chunkBytecode, getBlockIndex([x, y, z]));
 }
+
+export async function getBiome(
```
dhvanipa, line 157: moved this back here cuz it made more sense

packages/world/src/utils/OreLib.sol
```diff
@@ -47,191 +47,377 @@ library OreLib {
 
   function _getBiomeMultipliers(uint8 biomeIndex) public pure returns (uint256[6] memory) {
     // biomeIndex == 0: badlands
-    if (biomeIndex == 0) return [uint256(158630), 40481, 38193, 64501, 0, 10];
+    if (biomeIndex == 0) {
```
dhvanipa, line 50: previously we were only wrapping uint for the first value, changed it to all, not sure why we need it though? It was like this initially @vdrg

vdrg: it is so that the compiler can infer the type of the array

vdrg: yea

vdrg (approved):
nice!

PR #428, opened 2025-08-19
add player name system

frolic:


packages/world/src/systems/NameSystem.sol
```diff
@@ -0,0 +1,27 @@
...
+contract NameSystem is System {
+  function setPlayerName(EntityId caller, bytes32 name) public {
+    caller.activate();
+    address player = caller.getPlayerAddress();
+
+    require(ReversePlayerName._get(name) == address(0), "Player name is already in use.");
+
+    // free up previous name
+    // TODO: do we want to allow players to be able to change their name? if not, change this to a require/revert
+    bytes32 previousName = PlayerName._get(player);
```
dhvanipa, line 19: yea I think it makes sense to allow players to change their name

PR #432, opened 2025-08-25
Comment unused object names

vdrg:


vdrg:
closing as this doesn't seem to be the issue preventing the client from working locally

